PR #284, opened 2026-05-08
fix: emit &lt;lastmod&gt; in template sitemap (#283)

### LINKED ISSUE #283: Template's @astrojs/sitemap config emits no <lastmod>
The Anglesite template's `astro.config.ts` calls `sitemap()` with no options, so the generated sitemap has no `<lastmod>` element on any URL. The bundled `validateSitemap` flags this as a nice-to-have ("search engines use these to prioritize recrawling"), but the template's own default never satisfies its own audit.

A `serialize` hook is enough to emit a meaningful value:

```ts
sitemap({
  serialize(item) {
    return { ...item, lastmod: new Date().toISOString() };
  },
}),
```

Build-time stamp is approximate (every page looks like it changed on every deploy), but search engines tolerate it and it's strictly better than nothing. A more accurate version could read `dist/<path>/index.html` mtimes inside `serialize`, though that's noisier and harder to reason about for static-rebuild-on-every-deploy hosting like Cloudflare Pages/Workers.

## Suggested direction

Add the `serialize` hook to the template's `astro.config.ts` so freshly scaffolded sites ship with `<lastmod>` on every entry.

## Environment

- ANGLESITE_VERSION: 1.0.1
- Node: v22.19.0
- OS: darwin

davidwkeith:
## Summary

Fixes #283. The template's `astro.config.ts` was calling `sitemap()` with no options, so the generated sitemap had no `<lastmod>` element on any URL — and `validateSitemap` in the same template flagged its own default with `sitemap-no-lastmod`.

Pass a `serialize` hook that stamps each entry with the build-time ISO timestamp:

```ts
sitemap({
  serialize(item) {
    return { ...item, lastmod: new Date().toISOString() };
  },
}),
```

Build-time stamping is approximate (every page looks updated on every deploy) but search engines tolerate it, and it's strictly better than nothing. Freshly scaffolded sites now ship with `<lastmod>` on every entry.

## Test plan

- [ ] `npm run build` in a scaffolded site succeeds
- [ ] `dist/sitemap-0.xml` contains `<lastmod>` on every `<url>`
- [ ] `validateSitemap` no longer reports `sitemap-no-lastmod` on the default scaffold

https://claude.ai/code/session_014amXrMvG7ypHH4KGgrBkJh

---
_Generated by [Claude Code](https://claude.ai/code/session_014amXrMvG7ypHH4KGgrBkJh)_

### DIFF
--- a/template/astro.config.ts
+++ b/template/astro.config.ts
@@ -72,7 +72,11 @@ export default defineConfig({
     react(),
     markdoc(),
     ...(isDev ? [keystatic(), anglesiteToolbar()] : []),
-    sitemap(),
+    sitemap({
+      serialize(item) {
+        return { ...item, lastmod: new Date().toISOString() };
+      },
+    }),
   ],
   vite: { server: { https: getHttpsConfig() } },
   adapter: cloudflare(),
--- a/template/astro.config.ts
+++ b/template/astro.config.ts
@@ -18,6 +18,7 @@ import cloudflare from "@astrojs/cloudflare";
 import anglesiteToolbar from "./src/integrations/anglesite-toolbar";
 import { readFileSync, existsSync } from "node:fs";
 import { resolve } from "node:path";
+import { execSync } from "node:child_process";
 
 /** True when running `astro dev`, false during `astro build`. */
 const isDev =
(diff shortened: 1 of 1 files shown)


template/astro.config.ts
```diff
@@ -72,7 +72,11 @@ export default defineConfig({
     react(),
     markdoc(),
     ...(isDev ? [keystatic(), anglesiteToolbar()] : []),
-    sitemap(),
+    sitemap({
+      serialize(item) {
+        return { ...item, lastmod: new Date().toISOString() };
```
davidwkeith: What if the item already has a lastmod date? Could we use something more precise than the build date?

davidwkeith: Good catches — addressed in ac67aba:

1. **Preserve existing `lastmod`**: the serializer now early-returns `item` unchanged if `item.lastmod` is already set, so anything an upstream integration (or future Anglesite tweak) sets sticks.
2. **More precise than build date**: it now uses the source file's last-commit ISO timestamp from a one-shot `git log` cache. URL→source mapping tries `src/pages/<path>.{astro,md,mdoc,mdx,html}`, `src/pages/<path>/index.*`, and `src/content/<path>.*`. Build time is the fallback for unmatched URLs and environments without git history (shallow clones, non-repo builds).

---

PR #356, opened 2026-06-12
feat(server): list_content, create_page, create_post MCP tools (#140)

davidwkeith:
Implements **Anglesite-app#140** (Siri AI Phase A — A.6): the three plugin-side MCP tools the app's Phase A work depends on. Paired with Anglesite-app#171 (A.8), which consumes `list_content`.

## Tools

- **`list_content`** (read-only) — scans `src/pages/`, the article-like content collections (`posts`, `notes`, `episodes`, `experiments`) under `src/content/`, and `public/images/`, returning a **site-agnostic** JSON projection. No siteID, no entity ids — those are the app's to stamp.
- **`create_page`** — scaffolds a `BaseLayout` `.astro` page under `src/pages/`. Route and the relative layout-import depth are derived from the name (or an explicit `route`). Refuses to overwrite; won't scaffold the site root.
- **`create_post`** — scaffolds a **draft** content entry (default collection `posts` — the on-disk Astro collection name; the issue's colloquial "blog" maps here). Slug derived from the title. Refuses to overwrite.

## Contract with the app

`list_content`'s output matches the app's Swift `ContentListing.parse` (A.8) field-for-field, so it populates `SiteContentGraph` with no remapping:

| array | fields |
|---|---|
| `pages` | `route, filePath, title\|null, lastModified` |
| `posts` | `collection, slug, title, draft, publishDate\|null, tags, filePath, lastModified` |
| `images` | `relativePath, fileName, byteSize\|null, usedOnPages, lastModified` |

All dates are ISO strings (covered by the app's fractional-second decoder). `usedOnPages` is `[]` for now (reverse image→page lookup is an expensive cross-scan, deferred).

### Collection scope

Only article-like collections (`posts`/`notes`/`episodes`/`experiments`) map to `posts` — they carry title/publishDate/draft/tags. Gallery, team, menus, submissions, etc. don't fit the Post shape and would pollute Siri's "show my drafts", so they're excluded by design.

## Safety / honesty

- **Path traversal:** `create_post`'s `collection` is validated as a single safe segment (`^[A-Za-z0-9_-]+$`) so it can't escape `src/content/`; `create_page`'s `route` is fully slugified per segment.
- **Best-effort commit:** both create tools commit a single pathspec to HEAD and return the SHA, or `null` when there's no git repo / a hook rejects — the file still lands on disk (the source of truth). Unrelated staged/working changes are left untouched.
- Overwrite attempts throw → `{isError:true}` through the handler.

## Frontmatter

A minimal, dependency-free reader (`server/content-frontmatter.mjs`) parses the scalar/boolean/string-array subset the article collections actually use (quoted/unquoted scalars, booleans, inline `[a,b]` and block `- a` arrays). Deliberately not a general YAML parser.

## Tests

- `test/list-content.test.js` (10) — route derivation, title heuristic, dynamic-route/endpoint skipping, frontmatter fields, slug fallback, non-article exclusion, image scan, empty site.
- `test/create-content.test.js` (8) — scaffold + commit, nested import depth, no-overwrite, non-git fallback, slug/collection handling, **path-traversal rejection**.
- `tests/mcp-server.test.ts` — updated `tools/list` assertion + stdio e2e for `list_content` and `create_page`.

Full suite green apart from a **pre-existing** `optimize-images-packaging` failure (tsx sandbox path resolution; reproduces on `main`, unrelated to this change).

## Release coordination

Per the two-repo workflow, this ships in a tagged plugin release; the app PR then bumps the bundled-plugin pointer. CHANGELOG/version bump deferred to that release tagging (the CHANGELOG is backfilled per release, not per PR). The app degrades gracefully against older plugins without `list_content` (Anglesite-app#171), so ordering is safe.

### DIFF
--- a/server/create-content.mjs
+++ b/server/create-content.mjs
@@ -1,0 +1,166 @@
+/**
+ * `create_page` / `create_post` MCP tool backends (#140 / A.6).
+ *
+ * Both scaffold a minimal, valid source file from a fixed template and commit it onto the
+ * project's current branch (best-effort — the filesystem is the source of truth, so a missing
+ * git repo or a rejecting hook leaves the file on disk and reports `commit: null` rather than
+ * failing). Neither overwrites an existing file. These back the App Intents Add-Page / Add-Post
+ * flows (A.5) and feed the same content graph `list_content` reports.
+ *
+ * @module
+ */
+import { mkdirSync, writeFileSync, existsSync } from "node:fs";
+import { join, dirname } from "node:path";
+import { execFileSync } from "node:child_process";
+
+/**
+ * Scaffold a new Astro page under `src/pages/`.
+ *
+ * @param {string} projectRoot
+ * @param {{ name: string, route?: string }} input
+ * @returns {{ filePath: string, route: string, commit: string | null }}
+ */
+export function createPage(projectRoot, { name, route }) {
+  const title = (name ?? "").trim();
+  if (!title) throw new Error("create_page requires a non-empty name");
+
+  const normalizedRoute = normalizeRoute(route ?? slugify(title));
+  if (normalizedRoute === "/") {
+    throw new Error("create_page can't scaffold the site root; give the page a name or route");
+  }
+  const relPath = "src/pages" + normalizedRoute + ".astro";
+  const abs = join(projectRoot, relPath);
+  if (existsSync(abs)) throw new Error(`A page already exists at ${relPath}`);
+
+  // Depth below src/pages/ decides how many `../` the layout import needs.
+  const depth = normalizedRoute.replace(/^\//, "").split("/").length; // 1 for "/about"
+  const layoutImport = "../".repeat(depth) + "layouts/BaseLayout.astro";
+
+  writeFile(abs, pageTemplate({ title, layoutImport }));
+  return { filePath: relPath, route: normalizedRoute, commit: commitFile(projectRoot, relPath, `anglesite: add page ${normalizedRoute}`) };
+}
+
+/**
+ * Scaffold a new content-collection entry. Defaults to the `posts` collection (the on-disk
+ * Astro collection name; the issue's colloquial "blog" maps here). New posts are created as
+ * drafts so they stay out of the production build until the owner publishes.
+ *
+ * @param {string} projectRoot
+ * @param {{ title: string, collection?: string, slug?: string }} input
+ * @returns {{ filePath: string, slug: string, collection: string, commit: string | null }}
+ */
+export function createPost(projectRoot, { title, collection, slug }) {
+  const cleanTitle = (title ?? "").trim();
+  if (!cleanTitle) throw new Error("create_post requires a non-empty title");
+
+  const coll = (collection ?? "posts").trim() || "posts";
+  // `collection` becomes a path segment under src/content/, so it must not escape it. `slug`
+  // is neutralized by slugify below, but `collection` is used verbatim — restrict it to a
+  // single safe segment (no separators, dots, or traversal) rather than silently retargeting.
+  if (!/^[A-Za-z0-9_-]+$/.test(coll)) {
+    throw new Error(`Invalid collection name: ${coll}`);
+  }
+  const finalSlug = slugify(slug ?? cleanTitle);
+  if (!finalSlug) throw new Error("create_post could not derive a slug from the title");
+
+  const relPath = `src/content/${coll}/${finalSlug}.md`;
+  const abs = join(projectRoot, relPath);
+  if (existsSync(abs)) throw new Error(`A ${coll} entry already exists at ${relPath}`);
+
+  writeFile(abs, postTemplate({ title: cleanTitle }));
+  return {
+    filePath: relPath,
+    slug: finalSlug,
+    collection: coll,
+    commit: commitFile(projectRoot, relPath, `anglesite: add ${coll} ${finalSlug}`),
+  };
+}
+
+// MARK: - Templates
+
+function pageTemplate({ title, layoutImport }) {
+  const description = `${title}.`;
+  return `---
+import BaseLayout from "${layoutImport}";
+---
+
+<BaseLayout title="${escapeAttr(title)}" description="${escapeAttr(description)}">
+  <main>
+    <h1>${escapeHtml(title)}</h1>
+    <p>Add your content here.</p>
+  </main>
+</BaseLayout>
+`;
+}
+
+function postTemplate({ title }) {
+  const publishDate = new Date().toISOString();
+  return `---
+title: "${escapeYaml(title)}"
+description: ""
+publishDate: ${publishDate}
+draft: true
+tags: []
+---
+
+Write your post here.
+`;
+}
+
+// MARK: - Git (best-effort, commits to HEAD on the current branch)
+
+/**
+ * Stage and commit exactly `relPath` on the current branch. Returns the new HEAD SHA, or null
+ * on any failure (not a git repo, no prior commit, a pre-commit hook rejecting, git missing).
+ * Commits a single pathspec so unrelated staged/working changes are left untouched.
+ */
+function commitFile(projectRoot, relPath, message) {
+  const run = (args) =>
+    execFileSync("git", args, { cwd: projectRoot, encoding: "utf-8", stdio: ["ignore", "pipe", "pipe"] }).trim();
+  try {
+    run(["rev-parse", "--git-dir"]);
+    run(["add", "--", relPath]);
+    run(["commit", "-m", message, "--", relPath]);
+    return run(["rev-parse", "HEAD"]);
+  } catch {
+    return null;
+  }
+}
+
+// MARK: - Helpers
+
+function writeFile(abs, contents) {
+  mkdirSync(dirname(abs), { recursive: true });
+  writeFileSync(abs, contents);
+}
+
+/** `/About//Us/` → `/about-us`; bare name → `/name`. Always leading-slash, no trailing slash. */
+function normalizeRoute(route) {
+  const segments = String(route)
+    .split("/")
+    .map((s) => slugify(s))
+    .filter(Boolean);
+  return "/" + segments.join("/");
+}
+
+function slugify(value) {
+  return String(value)
+    .toLowerCase()
+    .normalize("NFKD")
+    .replace(/[̀-ͯ]/g, "") // strip combining diacritical marks
+    .replace(/['"]/g, "")
+    .replace(/[^a-z0-9]+/g, "-")
+    .replace(/^-+|-+$/g, "");
+}
+
+function escapeAttr(s) {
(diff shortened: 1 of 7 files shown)


server/create-content.mjs
```diff
@@ -0,0 +1,166 @@
...
+  const relPath = "src/pages" + normalizedRoute + ".astro";
+  const abs = join(projectRoot, relPath);
+  if (existsSync(abs)) throw new Error(`A page already exists at ${relPath}`);
+
+  // Depth below src/pages/ decides how many `../` the layout import needs.
+  const depth = normalizedRoute.replace(/^\//, "").split("/").length; // 1 for "/about"
+  const layoutImport = "../".repeat(depth) + "layouts/BaseLayout.astro";
+
+  writeFile(abs, pageTemplate({ title, layoutImport }));
+  return { filePath: relPath, route: normalizedRoute, commit: commitFile(projectRoot, relPath, `anglesite: add page ${normalizedRoute}`) };
```
davidwkeith, line 40: This line is very long and calls `commitFile` inline in the return statement. It's hard to scan. Consider breaking the call out:

```js
const commit = commitFile(projectRoot, relPath, `anglesite: add page ${normalizedRoute}`);
return { filePath: relPath, route: normalizedRoute, commit };
```

---

server/create-content.mjs
```diff
@@ -0,0 +1,166 @@
...
+  <main>
+    <h1>${escapeHtml(title)}</h1>
+    <p>Add your content here.</p>
+  </main>
+</BaseLayout>
+`;
+}
+
+function postTemplate({ title }) {
+  const publishDate = new Date().toISOString();
```
davidwkeith, line 97: Setting `publishDate` to the current timestamp for a draft post is misleading — the creation time is almost never the intended publish date. When the owner publishes later, this value will already be stale and will likely confuse the Siri "when did I publish this?" graph.

Consider omitting the field entirely for drafts (let the owner fill it in when publishing), or at least using a clearly-labelled placeholder date like `2000-01-01` that signals "not yet set".

---

server/create-content.mjs
```diff
@@ -0,0 +1,166 @@
...
+    .map((s) => slugify(s))
+    .filter(Boolean);
+  return "/" + segments.join("/");
+}
+
+function slugify(value) {
+  return String(value)
+    .toLowerCase()
+    .normalize("NFKD")
+    .replace(/[̀-ͯ]/g, "") // strip combining diacritical marks
```
davidwkeith, line 150: The code-point range `/[̀-ͯ]/` (U+0300–U+036F) covers the _Common Diacritical Marks_ block but misses many combining characters used in Arabic, Hebrew, Thai, Vietnamese, and emoji modifier sequences (U+0483–U+0489, U+0591–U+05C7, U+20D0–U+20FF, etc.). Titles in those scripts will slug-ify to an empty string, triggering the "could not derive a slug" error.

Node ≥ 22 supports Unicode property escapes, so this is a drop-in fix:

```js
.replace(/\p{M}/gu, "") // strip all Unicode combining marks
```

---

server/create-content.mjs
```diff
@@ -0,0 +1,166 @@
...
+    .replace(/['"]/g, "")
+    .replace(/[^a-z0-9]+/g, "-")
+    .replace(/^-+|-+$/g, "");
+}
+
+function escapeAttr(s) {
+  return String(s).replace(/&/g, "&amp;").replace(/"/g, "&quot;");
```
davidwkeith, line 157: `escapeAttr` only escapes `&` and `"`. It's used inside double-quoted attribute values, so `"` and `&` are the critical cases and this works for the current templates. However, if `>` appears in a title (e.g. `"A > B"`) it will land literally in the attribute value — technically valid HTML but surprising. A stricter escape avoids edge cases:

```js
return String(s).replace(/&/g, "&amp;").replace(/"/g, "&quot;").replace(/>/g, "&gt;");
```

---

server/list-content.mjs
```diff
@@ -0,0 +1,191 @@
...
+/** Raster/vector image extensions surfaced from `public/images/`. */
+const IMAGE_EXTENSIONS = new Set([".jpg", ".jpeg", ".png", ".webp", ".gif", ".svg", ".avif"]);
+/**
+ * Collections whose entries are "posts" in the graph's sense — they carry title / publishDate /
+ * draft / tags. Other collections (gallery, team, menus, submissions, …) don't fit the Post
+ * shape and would pollute Siri's "show my drafts", so they're intentionally excluded (#140).
+ */
+const ARTICLE_COLLECTIONS = new Set(["posts", "notes", "episodes", "experiments"]);
+
+/**
```
davidwkeith, line 32: `ARTICLE_COLLECTIONS` is hardcoded here and repeated verbatim in the `list_content` tool description in `index-tools.mjs`. A user who creates a custom collection called `articles` or `blog` will silently get nothing in `posts`. This will surface as a confusing Siri gap ("why can't Siri see my posts?").

Worth at minimum cross-referencing this with the Keystatic schema in the project (`keystatic.config.ts`), or documenting clearly in the tool description that only these four specific on-disk collection names are supported. If auto-detection is too complex for this PR, a fallback would be to check which of these directories actually exist before emitting empty results.

---

server/list-content.mjs
```diff
@@ -0,0 +1,191 @@
...
+  try {
+    text = readFileSync(abs, "utf-8");
+  } catch {
+    return null;
+  }
+  const fm = parseFrontmatter(text);
+  if (typeof fm.title === "string" && fm.title) return fm.title;
+  const m = /\btitle\s*=\s*(?:"([^"]*)"|'([^']*)')/.exec(text);
+  const found = m && (m[1] ?? m[2]);
+  return found ? found : null;
```
davidwkeith, line 90: The regex `/\btitle\s*=\s*(?:"([^"]*)"|'([^']*)')/` will match _any_ `title=` attribute in the file — ``, `<a title="tooltip">`, `<video title="…">`, etc. — not just the one on a layout component. For pages that have an `` near the top and no layout, this would return the image's alt-equivalent text as the page title.

A tighter heuristic would anchor on JSX-style component usage (uppercase-leading tag or known layout names):

```js
const m = /\b(?:BaseLayout|Layout)\b[^>]*\btitle\s*=\s*(?:"([^"]*)"|'([^']*)')/.exec(text);
```

Or just document the known limitation and let the app treat titles as untrusted/optional (which it already does via `title|null`).

---

server/index-tools.mjs
```diff
@@ -108,6 +110,59 @@ export function buildServer(projectRoot) {
...
   );
 
+  // Siri AI Phase A content tools (#140 / A.6). `list_content` feeds the Anglesite-app's
+  // SiteContentGraph (#142); `create_page`/`create_post` back the Add-Page/Add-Post intents (A.5).
+  server.tool(
+    "list_content",
+    "List the site's pages, article-like content-collection entries (posts, notes, episodes, experiments), and images under public/images, as structured JSON. Read-only; the filesystem is the source of truth.",
+    {},
+    () => {
+      const listing = listContent(projectRoot);
```
davidwkeith, line 120: `list_content` has no error handling, but `create_page` and `create_post` do. If `listContent` throws (e.g. a permission error on a file mid-scan, or an unexpected `statSync` failure) the exception will bubble uncaught into the MCP server framework. The `walk` helper handles missing directories gracefully, but individual `readFileSync`/`statSync` calls inside `pageTitle` and `scanImages` can still throw in edge cases.

For consistency, wrap this the same way as the create tools:

```js
try {
  const listing = listContent(projectRoot);
  return { content: [{ type: "text", text: JSON.stringify(listing) }] };
} catch (error) {
  return { content: [{ type: "text", text: error.message }], isError: true };
}
```

---

test/create-content.test.js
```diff
@@ -0,0 +1,110 @@
...
+beforeEach(initRepo);
+afterEach(() => repo && rmSync(repo, { recursive: true, force: true }));
+
+describe("createPage", () => {
+  it("scaffolds a top-level page, derives the route from the name, and commits", () => {
+    const result = createPage(repo, { name: "About Us" });
+
+    expect(result.route).toBe("/about-us");
+    expect(result.filePath).toBe("src/pages/about-us.astro");
+    expect(result.commit).toMatch(/^[0-9a-f]{40}$/);
```
davidwkeith, line 37: This regex hardcodes a 40-character SHA-1 hash. Git ≥ 2.42 supports SHA-256 repositories (64 hex chars), and while SHA-1 is still the default today, this test will break if the CI runner ever initialises a SHA-256 repo. A more resilient pattern:

```js
expect(result.commit).toMatch(/^[0-9a-f]{40,64}$/);
```

---

davidwkeith (commented):
## Overview

This is a well-structured PR. The separation of concerns across `content-frontmatter.mjs`, `list-content.mjs`, and `create-content.mjs` is clean, the "best-effort commit" design is pragmatic, the path-traversal rejection is solid, and the test coverage is good. The issues below are mostly correctness bugs rather than architectural concerns.

## Issues requiring fixes

**`slugify` — incomplete diacritical mark range (bug)**
The regex `/[̀-ͯ]/` only covers U+0300–U+036F. Combining marks in Arabic, Hebrew, Thai, and other scripts fall outside that range, so non-Latin titles will slugify to an empty string and throw. Replace with `/\p{M}/gu` (Node ≥ 22 supports this natively). See inline comment on `create-content.mjs:150`.

**`publishDate` set for draft posts (design bug)**
Draft posts scaffold with `publishDate` set to the moment of creation. When the owner publishes weeks later, that timestamp will already be wrong, and the Siri content graph will have a stale date baked in. Omitting the field for drafts is the safer default. See inline comment on `create-content.mjs:97`.

**`pageTitle` regex too broad (reliability bug)**
The title extraction regex matches any `title=` attribute in the file — ``, `<a>`, etc. — not just layout component props. Pages with non-layout `title=` attributes near the top will return a misleading title. See inline comment on `list-content.mjs:90`.

**`list_content` handler has no error handling (inconsistency)**
`create_page` and `create_post` both catch errors and return `{isError:true}`. `list_content` doesn't. An unexpected `readFileSync`/`statSync` failure mid-scan will bubble uncaught. See inline comment on `index-tools.mjs`.

## Minor issues

- **`escapeAttr` missing `>`** — doesn't matter for current templates but is a footgun for future template changes. (inline comment on `create-content.mjs:157`)
- **Long return line in `createPage`** — the `commitFile` call inlined into the `return` makes the line excessively long. (inline comment on `create-content.mjs:40`)
- **SHA regex in tests** — hardcoded 40-char SHA-1 pattern; use `{40,64}` to survive SHA-256 repos. (inline comment on `create-content.test.js:37`)
- **Hardcoded `ARTICLE_COLLECTIONS`** — a user with a `articles` or `blog` collection will silently get no posts. Worth documenting clearly in the tool description even if auto-detection is deferred. (inline comment on `list-content.mjs:32`)


---

Issue #150, opened 2026-03-30
ADR-0015 (site search) still in proposed status


davidwkeith:
## Problem

ADR-0015 "site-search" is still marked `status: proposed`, while the `search` skill already ships Pagefind as the chosen solution.

## Suggested fix

Accept ADR-0015 — the decision has effectively been made and implemented. Update the status field from `proposed` to `accepted`.

## Priority

Should fix before 1.0 — all ADRs should reflect actual decisions.

PR #359, opened 2026-06-12
feat(server): add list_content MCP tool (pages/posts/images)

### LINKED ISSUE #358: server: add list_content MCP tool (pages/posts/images) — unblocks app content graph + Siri typed entities
## Summary

Add a `list_content` tool to the MCP server (`server/index-tools.mjs`) that returns a site's pages, posts, and images. The **Anglesite-app already consumes this tool** — it's the plugin half of a paired feature. Until it ships, the app's content graph stays empty and Siri's typed-entity resolution can't fire.

## Why

The app's `SiteContentGraph` is populated by calling `list_content` over MCP on every site open (`Anglesite-app` → `Sources/AnglesiteCore/LocalSiteRuntime.swift:165`, `populateContentGraph`). The server currently registers only `add_annotation`, `list_annotations`, `resolve_annotation`, `apply_edit`, `undo_edit` — no `list_content` — so the call fails "tool not found" and the app logs *"list_content not available (older plugin) — content graph left empty"* and continues.

Consequence in the app: every previewed element (including `` and posts) falls through to a generic transient `ElementEntity`. The typed `PageEntity` / `PostEntity` / `ImageEntity` resolution that powers Siri onscreen-awareness (Anglesite-app Phase A #132 / Phase B #133, smoke #150) never runs. This issue is the unblocker.

## Wire contract (already defined app-side)

The app decodes the response with `ContentListing.parse` (`Sources/AnglesiteCore/ContentListing.swift`). Match this shape exactly. Return it the same way the other tools do — a single text-content item whose `text` is the JSON below:

```js
return { content: [{ type: "text", text: JSON.stringify(payload) }] };
```

```jsonc
{
  "pages": [
    {
      "route": "/about",                 // required — URL route
      "filePath": "src/pages/about.astro",// required — projectRoot-relative
      "title": "About",                  // optional
      "lastModified": "2026-06-12T18:30:00Z" // required — ISO-8601 (plain or fractional)
    }
  ],
  "posts": [
    {
      "collection": "blog",              // required
      "slug": "hello-anglesite",         // required
      "title": "Hello, Anglesite",       // required
      "draft": false,                    // required
      "publishDate": "2026-06-12T00:00:00Z", // optional
      "tags": ["smoke", "siri"],         // required (may be [])
      "filePath": "src/content/blog/hello-anglesite.md", // required
      "lastModified": "2026-06-12T18:30:00Z" // required
    }
  ],
  "images": [
    {
      "relativePath": "public/images/hero.svg", // required — projectRoot-relative, no leading slash
      "fileName": "hero.svg",            // required — last path component
      "byteSize": 1234,                  // optional (int)
      "usedOnPages": ["/", "/blog/hello-anglesite"], // required (may be [])
      "lastModified": "2026-06-12T18:30:00Z" // required
    }
  ]
}
```

Contract notes (all enforced by the app decoder):
- **Site-agnostic payload.** No `siteID`, no entity `id` — the app stamps `{siteID}:page:{route}` / `:post:{slug}` / `:image:{relativePath}` itself. Don't add them.
- **Top-level arrays are optional.** A missing `pages`/`posts`/`images` key decodes to `[]` — fine to omit a section you don't yet discover.
- **Timestamps are ISO-8601**, plain (`…Z`) or fractional (`….500Z`). `lastModified` required on every entry; `publishDate` optional (posts only).
- **`images.relativePath` has no leading slash** (e.g. `public/images/hero.svg`). The app's `imageMatches` does a path-suffix + filename-fallback match against rendered ``, so `relativePath` + `fileName` must be accurate.

## Discovery (suggested)

- **pages** — `src/pages/**` routable files (`.astro`/`.md`/`.mdx`), route derived the way Astro derives it; `title` from frontmatter when present.
- **posts** — Astro content collections (`src/content/<collection>/**` or the project's blog convention); `slug`, `draft`, `publishDate`, `tags` from frontmatter.
- **images** — `public/**` and `src/assets/**` raster/vector assets. `usedOnPages` is best-effort (scan rendered/source references); `[]` is acceptable for v1.

## Acceptance criteria

- [ ] `list_content` registered in `server/index-tools.mjs` with a Zod input schema (no required args; consider an optional `kinds` filter later).
- [ ] Returns the JSON shape above as a single text-content item; `isError: true` on failure (mirrors `resolve_annotation`).
- [ ] Tests in `test/` covering a fixture site with ≥1 page, ≥1 post (incl. a draft), ≥1 image with `usedOnPages`.
- [ ] Manual: with the matching app build, open a site → app log shows the graph populated (not "content graph left empty"); a previewed `` resolves to `ImageEntity` and a post element with `data-anglesite-id` resolves to `PostEntity`.

## Cross-refs

- App consumer: `Anglesite-app` `LocalSiteRuntime.swift:165` (`populateContentGraph`), decoder `ContentListing.swift`.
- App epics this unblocks: Anglesite-app #132 (Phase A entities), #133 (Phase B onscreen awareness), #150 (manual smoke).
- Two-repo flow (per app `CLAUDE.md`): this plugin PR ships in a tagged plugin release; the app then bumps its bundled-plugin pointer.

davidwkeith:
## Summary

Adds a `list_content` MCP tool that returns a site's **pages, posts, and images** as a site-agnostic JSON snapshot. This is the **plugin half of a paired feature** (per the two-repo flow): the Anglesite-app already calls `list_content` to populate its `SiteContentGraph` and currently logs *"list_content not available (older plugin) — content graph left empty"*, so Siri's typed-entity resolution (`PageEntity`/`PostEntity`/`ImageEntity`) never fires. This tool is the unblocker.

## What it discovers

- **pages** — `src/pages/**`, Astro route derivation (`index` → `/`, dynamic `[slug]`/`[...rest]` routes skipped); `title` from frontmatter.
- **posts** — `src/content/<collection>/**` (`.md`/`.mdx`/`.mdoc`): `collection`, `slug` (glob-loader id), `title`, `draft`, `tags`; `publishDate` **normalized to full ISO-8601** (the app's `ISO8601DateFormatter` rejects date-only strings).
- **images** — `public/**` + `src/assets/**`: `relativePath` (no leading slash), `fileName`, `byteSize`, `lastModified`, best-effort `usedOnPages`.

Payload is site-agnostic — no `siteID`, no entity ids; the host stamps those (`Sources/AnglesiteCore/ContentListing.swift`). Frontmatter is parsed with a small no-dependency helper (repo ships no YAML lib).

## Tests (TDD)

`test/list-content.test.js` — 6 tests, each written failing-first:
- pages discovery + route derivation; dynamic-route skip
- posts: fields, defaults (`draft`/`tags`), `publishDate` ISO-8601 normalization
- images: metadata; best-effort `usedOnPages`
- `list_content` MCP-tool registration + return shape (in-memory client/server)

Also updated `tests/mcp-server.test.ts`'s exact tool-name set to include `list_content`.

**Verification:** all 6 new tests green; real end-to-end run against a seeded site produced correct pages + image with accurate `usedOnPages`. The 5 unrelated failing suites locally are the pre-existing `sharp`-not-installed gap (module-load errors in `optimize-images.mjs`, untouched here) — they pass in CI.

## Follow-up (app side)

Bump the app's bundled-plugin pointer once this ships in a tagged plugin release, then the app's content graph populates and `ImageEntity`/`PostEntity` resolution comes alive for the #150 smoke.

Closes #358

### DIFF
--- a/server/list-content.mjs
+++ b/server/list-content.mjs
@@ -1,0 +1,242 @@
+import { readdirSync, readFileSync, statSync } from "node:fs";
+import { join, relative, sep } from "node:path";
+
+/**
+ * Discover a site's content (pages, posts, images) for the `list_content` MCP tool.
+ *
+ * The payload is **site-agnostic**: no `siteID`, no entity `id` — the Anglesite-app
+ * stamps those itself (`Sources/AnglesiteCore/ContentListing.swift`). Timestamps are
+ * ISO-8601; `filePath`/`relativePath` are projectRoot-relative with POSIX separators.
+ *
+ * @param {string} projectRoot Absolute path to the site root.
+ * @returns {{ pages: object[], posts: object[], images: object[] }}
+ */
+export function listContent(projectRoot) {
+  return {
+    pages: discoverPages(projectRoot),
+    posts: discoverPosts(projectRoot),
+    images: discoverImages(projectRoot),
+  };
+}
+
+// ── Pages ─────────────────────────────────────────────────────────
+
+const PAGE_EXTENSIONS = [".astro", ".md", ".mdx", ".html"];
+
+function discoverPages(projectRoot) {
+  const pagesDir = join(projectRoot, "src", "pages");
+  const files = walkFiles(pagesDir, (name) =>
+    PAGE_EXTENSIONS.some((ext) => name.endsWith(ext)),
+  );
+  const pages = [];
+  for (const file of files) {
+    const routePath = relative(pagesDir, file);
+    // Dynamic routes (`[slug]`, `[...rest]`) need params we can't enumerate
+    // statically — skip them rather than emit a literal `/blog/[slug]` route.
+    if (routePath.includes("[")) continue;
+    const page = {
+      route: deriveRoute(routePath),
+      filePath: toRelativePosix(projectRoot, file),
+      lastModified: mtimeISO(file),
+    };
+    const title = scalar(frontmatterBlock(readFileSafe(file)), "title");
+    if (title) page.title = title;
+    pages.push(page);
+  }
+  return pages;
+}
+
+/** `index.astro` → `/`, `about.astro` → `/about`, `blog/index.astro` → `/blog`. */
+function deriveRoute(routePath) {
+  let r = routePath.split(sep).join("/");
+  r = r.replace(/\.[^.]+$/, ""); // drop extension
+  r = r.replace(/(^|\/)index$/, ""); // collapse trailing index
+  return "/" + r;
+}
+
+// ── Images ────────────────────────────────────────────────────────
+
+const IMAGE_EXTENSIONS = [
+  ".png", ".jpg", ".jpeg", ".gif", ".svg", ".webp", ".avif", ".ico",
+];
+
+function discoverImages(projectRoot) {
+  const dirs = [join(projectRoot, "public"), join(projectRoot, "src", "assets")];
+  const isImage = (name) =>
+    IMAGE_EXTENSIONS.some((ext) => name.toLowerCase().endsWith(ext));
+
+  // Precompute (route, source) for every page once, so usedOnPages is a
+  // substring scan rather than a re-walk per image.
+  const pagesDir = join(projectRoot, "src", "pages");
+  const pageRefs = walkFiles(pagesDir, (name) =>
+    PAGE_EXTENSIONS.some((ext) => name.endsWith(ext)),
+  ).map((file) => ({
+    route: deriveRoute(relative(pagesDir, file)),
+    text: readFileSafe(file),
+  }));
+
+  const images = [];
+  for (const dir of dirs) {
+    for (const file of walkFiles(dir, isImage)) {
+      const fileName = file.split(sep).pop();
+      const usedOnPages = pageRefs
+        .filter((p) => p.text.includes(fileName))
+        .map((p) => p.route);
+      images.push({
+        relativePath: toRelativePosix(projectRoot, file),
+        fileName,
+        byteSize: statSync(file).size,
+        usedOnPages: [...new Set(usedOnPages)].sort(),
+        lastModified: mtimeISO(file),
+      });
+    }
+  }
+  return images;
+}
+
+// ── Posts ─────────────────────────────────────────────────────────
+
+const POST_EXTENSIONS = [".mdoc", ".mdx", ".md"];
+
+function discoverPosts(projectRoot) {
+  const contentDir = join(projectRoot, "src", "content");
+  let collections;
+  try {
+    collections = readdirSync(contentDir, { withFileTypes: true });
+  } catch {
+    return [];
+  }
+  const posts = [];
+  for (const entry of collections) {
+    if (!entry.isDirectory()) continue;
+    const collection = entry.name;
+    const collectionDir = join(contentDir, collection);
+    const files = walkFiles(collectionDir, (name) =>
+      POST_EXTENSIONS.some((ext) => name.endsWith(ext)),
+    );
+    for (const file of files) {
+      const slug = deriveSlug(relative(collectionDir, file));
+      const fm = frontmatterBlock(readFileSafe(file));
+      const post = {
+        collection,
+        slug,
+        title: scalar(fm, "title") ?? slug,
+        draft: boolean(fm, "draft") ?? false,
+        tags: stringArray(fm, "tags"),
+        filePath: toRelativePosix(projectRoot, file),
+        lastModified: mtimeISO(file),
+      };
+      const publishDate = normalizeDate(scalar(fm, "publishDate"));
+      if (publishDate) post.publishDate = publishDate;
+      posts.push(post);
+    }
+  }
+  return posts;
+}
+
+/** Astro glob-loader id: path under the collection base, minus extension. */
+function deriveSlug(entryPath) {
+  return entryPath.split(sep).join("/").replace(/\.[^.]+$/, "");
+}
+
+// ── Shared helpers ──────────────────────────────────────────────────
+
+/** Recursively collect files matching `predicate`. Skips build/vcs dirs. */
+function walkFiles(dir, predicate) {
+  const results = [];
+  let entries;
+  try {
+    entries = readdirSync(dir, { withFileTypes: true });
+  } catch {
+    return results;
+  }
+  for (const entry of entries) {
+    if (["node_modules", ".git", "dist", ".astro"].includes(entry.name)) continue;
+    const full = join(dir, entry.name);
+    if (entry.isDirectory()) {
+      results.push(...walkFiles(full, predicate));
+    } else if (predicate(entry.name)) {
+      results.push(full);
+    }
+  }
+  return results;
+}
+
+function toRelativePosix(projectRoot, file) {
+  return relative(projectRoot, file).split(sep).join("/");
+}
+
+function mtimeISO(file) {
+  return statSync(file).mtime.toISOString();
+}
+
+function readFileSafe(file) {
+  try {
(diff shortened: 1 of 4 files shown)


server/list-content.mjs
```diff
@@ -0,0 +1,242 @@
...
+}
+
+/** Normalize a date string to a full ISO-8601 datetime, or `undefined`. */
+function normalizeDate(raw) {
+  if (!raw) return undefined;
+  const d = new Date(raw);
+  return Number.isNaN(d.getTime()) ? undefined : d.toISOString();
+}
+
+function stripQuotes(v) {
```
davidwkeith, line 240: **Bug: mismatched quote stripping.**

The regex `/^["'](.*)["']$/` allows mixed delimiters — it matches `"hello'` or `'world"` as well as properly-quoted strings. Use a backreference so the same character must open and close:

```js
function stripQuotes(v) {
  return v.replace(/^(['"])(.*)\1$/, "$2");
}
```

In practice, well-formed YAML won't produce mismatched quotes, but this is the kind of silent mis-parse that's hard to debug when it eventually surfaces.

---

server/list-content.mjs
```diff
@@ -0,0 +1,242 @@
...
+  }
+}
+
+// ── Minimal frontmatter parsing ─────────────────────────────────────
+// The repo ships no YAML dependency; these cover the scalar/boolean/list
+// shapes the content schemas use (see template/src/content.config.ts).
+
+/** The raw text inside a leading `---` … `---` block, or "" if absent. */
+function frontmatterBlock(source) {
+  const m = source.match(/^---\n([\s\S]*?)\n---/);
```
davidwkeith, line 187: **CRLF line-endings break frontmatter parsing.**

The regex `^---\n` requires a bare `\n`, so any file with Windows CRLF line-endings (`---\r\n`) silently returns no frontmatter — all posts lose their `title`, `draft`, `tags`, and `publishDate`. This affects content files authored or committed on Windows, or repos with `core.autocrlf = true`.

```js
function frontmatterBlock(source) {
  const m = source.match(/^---\r?\n([\s\S]*?)\r?\n---/);
  return m ? m[1].replace(/\r/g, "") : "";
}
```

The `.replace(/\r/g, "")` on the captured block ensures the scalar/array regexes (which also use `\n` literally) continue to work after extraction.

---

server/list-content.mjs
```diff
@@ -0,0 +1,242 @@
...
+
+  // Precompute (route, source) for every page once, so usedOnPages is a
+  // substring scan rather than a re-walk per image.
+  const pagesDir = join(projectRoot, "src", "pages");
+  const pageRefs = walkFiles(pagesDir, (name) =>
+    PAGE_EXTENSIONS.some((ext) => name.endsWith(ext)),
+  ).map((file) => ({
+    route: deriveRoute(relative(pagesDir, file)),
+    text: readFileSafe(file),
+  }));
```
davidwkeith, line 76: **Unnecessary second filesystem walk.**

`discoverImages` re-walks `src/pages` and re-reads every page file from disk — a full duplicate of what `discoverPages` already did. On a large site this is a meaningful cost.

The cleanest fix is to compute `pageRefs` once in `listContent` and pass it to both callers:

```js
export function listContent(projectRoot) {
  const pagesDir = join(projectRoot, "src", "pages");
  const pageFiles = walkFiles(pagesDir, (name) =>
    PAGE_EXTENSIONS.some((ext) => name.endsWith(ext)),
  );
  const pageRefs = pageFiles.map((file) => ({
    route: deriveRoute(relative(pagesDir, file)),
    text: readFileSafe(file),
  }));
  return {
    pages: discoverPages(projectRoot, pagesDir, pageRefs),
    posts: discoverPosts(projectRoot),
    images: discoverImages(projectRoot, pageRefs),
  };
}
```

Then `discoverPages` and `discoverImages` each receive `pageRefs` instead of re-deriving it. One walk, one set of reads.

---

server/list-content.mjs
```diff
@@ -0,0 +1,242 @@
...
+
+/** The raw text inside a leading `---` … `---` block, or "" if absent. */
+function frontmatterBlock(source) {
+  const m = source.match(/^---\n([\s\S]*?)\n---/);
+  return m ? m[1] : "";
+}
+
+/** A single-line scalar value (`key: value`), quotes stripped. */
+function scalar(fm, key) {
+  const line = fm.match(new RegExp(`^${key}:[ \\t]*(.+)$`, "m"));
```
davidwkeith, line 193: **Regex key is not escaped.**

`key` is interpolated directly into `new RegExp(...)` here and in `stringArray` (line 214). Frontmatter keys are always alphanumeric in practice, but if a key ever contained a regex metacharacter it would produce a silently wrong match.

```js
const escaped = key.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
const line = fm.match(new RegExp(`^${escaped}:[ \\t]*(.+)$`, "m"));
```

Apply the same change to both `new RegExp` calls in `stringArray`.

---

server/list-content.mjs
```diff
@@ -0,0 +1,242 @@
...
+    route: deriveRoute(relative(pagesDir, file)),
+    text: readFileSafe(file),
+  }));
+
+  const images = [];
+  for (const dir of dirs) {
+    for (const file of walkFiles(dir, isImage)) {
+      const fileName = file.split(sep).pop();
+      const usedOnPages = pageRefs
+        .filter((p) => p.text.includes(fileName))
```
davidwkeith, line 83: **`usedOnPages` false-positives when two images share a filename.**

`p.text.includes(fileName)` uses only the bare filename (e.g. `hero.svg`). If `public/images/hero.svg` and `src/assets/hero.svg` both exist, each will claim the other's page references.

The PR description calls this "best-effort", which is fine, but a two-segment key cuts most false positives at no extra cost:

```js
// Use the last two path segments ("images/hero.svg") to reduce false matches.
const fileKey = toRelativePosix(projectRoot, file).split("/").slice(-2).join("/");
const usedOnPages = pageRefs
  .filter((p) => p.text.includes(fileKey))
  .map((p) => p.route);
```

---

davidwkeith (commented):
The structure is clean and the TDD approach is well-executed. Two actual bugs need fixing before merge; two more are latent but worth addressing now.

**Must fix**

- **`frontmatterBlock` CRLF** (line 187) — `^---\n` fails silently on any content file with Windows line-endings. A site cloned on Windows would produce a `list_content` response where every post loses its `title`, `draft`, `tags`, and `publishDate` with no error surfaced — the most impactful bug here.
- **`stripQuotes` mismatched quotes** (line 240) — the regex allows `"value'`; fix with a backreference `\1`.

**Should fix**

- **Double filesystem walk in `discoverImages`** (lines 71–76) — re-walks `src/pages` and re-reads every page file that `discoverPages` already processed. Refactor `listContent` to compute `pageRefs` once and pass it down to both callers.
- **Unescaped regex key in `scalar`/`stringArray`** (lines 193, 214) — `new RegExp(key)` without escaping; safe for today's alphanumeric keys but worth hardening.

**Note (no action required)**

- `usedOnPages` false-positives for same-named images in different directories — already described as "best-effort" in the PR. See inline for a simple two-segment mitigation.

---

davidwkeith:
Closing as a duplicate. `main` already shipped `list_content` (plus `create_page`/`create_post`) via #140/#356 (`c4275d3`) before this branch was opened — I missed the in-flight PR because I only dup-checked open issues, not merged work.

That implementation is also more correct than this one: it restricts posts to article collections (`posts, notes, episodes, experiments`) rather than treating every `src/content/*` dir as a post, and uses a real frontmatter parser. The only things this branch added that #356 lacks are best-effort `usedOnPages` and `src/assets/` image scanning — captured as a possible follow-up rather than carried here.

Issue #251, opened 2026-05-07
Use Cloudflare MCP accounts_list / set_active_account in start + deploy


davidwkeith:
## Problem

`skills/start/SKILL.md` and `skills/deploy/SKILL.md` rely on whatever account `wrangler whoami` happens to return. Owners who are logged into both a personal and a business Cloudflare account (very common — agencies, freelancers, side projects) can silently deploy to the wrong account with no in-flow confirmation.

## Proposal

Use Cloudflare MCP tools `accounts_list` and `set_active_account`:

1. In `start`, after Cloudflare login, call `accounts_list`. If more than one account is returned, present a numbered picker and `set_active_account`.
2. Save the chosen `CLOUDFLARE_ACCOUNT_ID` to `.site-config`.
3. In `deploy`, validate the active account matches `.site-config` before running `wrangler deploy`. If they diverge, abort with a clear message rather than silently deploying somewhere unexpected.

## Scope

- Update `skills/start/SKILL.md` — account selection step
- Update `skills/deploy/SKILL.md` — pre-deploy account validation
- Add `CLOUDFLARE_ACCOUNT_ID` documentation to `.site-config` schema
- Add `mcp__cloudflare__accounts_list` and `mcp__cloudflare__set_active_account` to `allowed-tools`
- Consider whether `check` should report current active account vs. configured account

PR #366, opened 2026-06-19
feat(apply-edit): dry_run preview + edit-style op (paired with app #251)

davidwkeith:
Plan A of the Anglesite-app **#251** epic (Siri natural-language edits with a reviewed before/after diff). This is the **plugin prerequisite**; the app PR (Plan B) follows after this release is tagged and bumps the bundled-plugin pointer.

## What this adds

**1. `apply_edit` `dry_run` mode (read-only preview)**
- New optional `dry_run: boolean`. When true, computes the would-be change and returns a new `anglesite:edit-preview` body `{ id, file, range, op, before, after }` **without writing to disk or committing history**.
- `before`/`after` are the spliced region plus a bounded ±200-char context window.
- Op-agnostic: text/attr/style all preview through the same short-circuit (after `spliceSource`, before `atomicWrite`).
- `replace-image-src` + `dry_run` refuses (`not-implemented`) **before** image processing, preserving the read-only invariant (image previews would require writing optimized bytes).
- Refusals (`no-match`/`ambiguous`/…) still return correctly under `dry_run`.

**2. New `edit-style` op**
- Value `{ property, value }`. Merges a CSS rule into the **owning Astro component's scoped `<style>`** (created if absent) — component-encapsulation as the CSS-location resolver.
- Targets the element by existing `#id`/`.class`; adds a deterministic marker class (`ang-<6hex>`) to the opening tag only when it has neither.
- Returns a whole-file replacement over `range {0, source.length}` (two disjoint edit regions), reusing the existing splice/write/dry_run path unchanged.
- Handles `<style lang=…>` / `<style is:global>` variants and dedups an already-present property (won't strip superset properties like `background-color` when setting `color`).

## Tests
22 feature tests across `apply-edit-preview`, `apply-edit-dry-run`, `style-edit`, `patcher-edit-style`, and `mcp-server` (stdio e2e). Read-only dry_run verified by byte-identical-file assertions at unit and e2e levels. (Two pre-existing, unrelated suite failures — `optimize-images-packaging`/sharp and an `update.test.ts` shell timeout — are not touched by this branch.)

## Release sequencing
Merge this, then tag **v1.1.0** (version bumped in `package.json` + `.claude-plugin/plugin.json` + `CLAUDE.md`) to publish the artifact the app's Plan B consumes.

## Follow-up (non-blocking)
Consider a `dry-run-unsupported` refusal reason (clearer than `not-implemented`) for image dry_run.

Design + plan: `Anglesite-app/docs/specs/2026-06-19-siri-nl-edit-{interpretation-design,plugin-plan}.md`.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "anglesite",
-  "version": "1.0.2",
+  "version": "1.1.0",
   "description": "AI webmaster that scaffolds, designs, and deploys Astro + Keystatic sites on Cloudflare Workers.",
   "author": {
     "name": "David W. Keith"
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -2,7 +2,7 @@
 
 Anglesite is a Claude plugin that scaffolds and manages websites for small businesses. It works with Claude Cowork (non-technical site owners, GUI) and Claude Code (developers, CLI). It generates Astro + Keystatic sites deployed to Cloudflare Workers (Static Assets, via the `@astrojs/cloudflare` adapter).
 
-**Version:** 1.0.2 · **License:** ISC · **Node:** >=22 · **Module system:** ESM
+**Version:** 1.1.0 · **License:** ISC · **Node:** >=22 · **Module system:** ESM
 
 ## Plugin structure
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "anglesite",
-  "version": "1.0.2",
+  "version": "1.1.0",
   "type": "module",
   "description": "AI webmaster that scaffolds, designs, and deploys Astro + Keystatic sites on Cloudflare Workers.",
   "scripts": {
--- a/server/apply-edit-dispatcher.mjs
+++ b/server/apply-edit-dispatcher.mjs
@@ -35,6 +35,7 @@ import { optimizeImage } from "./optimize-images.mjs";
 import {
   createEditAppliedContent,
   createEditFailedContent,
+  createEditPreviewContent,
 } from "./apply-edit-schema.mjs";
 
 function failed(id, reason, detail) {
@@ -48,6 +49,19 @@ function applied(id, file, range, commit, result) {
 /** Splice `replacement` into `source` at the resolved byte range. */
 function spliceSource(source, range, replacement) {
   return source.slice(0, range.start) + replacement + source.slice(range.end);
+}
+
+/** Bounded before/after fragments around a [start,end) splice — keeps preview payloads small. */
+function windowAround(source, start, end, replacement, pad = 200) {
+  const from = Math.max(0, start - pad);
+  const to = Math.min(source.length, end + pad);
+  const before = source.slice(from, to);
+  const after = source.slice(from, start) + replacement + source.slice(end, to);
+  return { before, after };
+}
+
+function preview(id, file, range, op, before, after) {
+  return { content: [createEditPreviewContent(id, file, range, op, before, after)] };
 }
 
 /** Atomic write via temp-sibling + rename, so a crashed mid-write can't truncate the source. */
@@ -195,6 +209,11 @@ export async function applyEdit(projectRoot, edit, opts = {}) {
   }
 
   const next = spliceSource(source, range, replacement);
+
+  if (edit.dry_run) {
+    const { before, after } = windowAround(source, range.start, range.end, replacement);
+    return preview(edit.id, file, range, edit.op, before, after);
+  }
 
   try {
     atomicWrite(absPath, next);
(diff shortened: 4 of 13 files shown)


davidwkeith (commented):
Good shape overall — the `dry_run` short-circuit is clean, the `edit-style` op is well-motivated, and the test coverage is thorough. Four real issues need fixing before this merges, plus two minor ones.

---

## Bugs

### 1. `windowAround` sends the entire file for `edit-style` (defeats the payload limit)

`resolveStyle` in `patcher.mjs` returns `range: { start: 0, end: source.length }` (a whole-file replacement). When that range is passed to `windowAround` in `apply-edit-dispatcher.mjs`:

```js
const from = Math.max(0, 0 - 200);           // → 0
const to   = Math.min(source.length, source.length + 200); // → source.length
const before = source.slice(0, source.length); // = entire file
const after  = source.slice(0, 0) + replacement + source.slice(source.length, source.length); // = replacement
```

The ±200-char padding has no effect when `start === 0` and `end === source.length`. For a 500-line Astro component the dry-run preview sends the whole file twice. The docstring says "keeps preview payloads small" — that invariant is broken for `edit-style`.

**Fix**: For whole-file ranges, clip the preview around only the lines that changed (e.g. find the first differing character and apply the pad there), or document that `edit-style` dry-runs return a full-file diff and accept that tradeoff explicitly.

---

### 2. `value.value` is not validated — produces `color: undefined;`

`resolveStyle` in `patcher.mjs` guards against a missing `value.property` but not `value.value`:

```js
if (!value || typeof value !== "object" || !value.property) {
  return refuse("no-match", "edit-style value must be { property, value }");
}
```

A call with `{ property: "color" }` (no `value` key) passes this check, proceeds through `rewriteAstroStyle(source, selector, "color", undefined)`, and writes `color: undefined;` to disk. No test covers this.

**Fix**: Add `|| value.value === undefined` to the guard, and add a test.

---

### 3. `template/package.json` not bumped — version sync violation

`CLAUDE.md` is explicit: *"Versions must stay in sync across three files: `package.json`, `.claude-plugin/plugin.json`, `template/package.json`."* The diff bumps the first two from `1.0.2` to `1.1.0` but leaves `template/package.json` untouched. The release CI step verifies this and will fail on tag.

---

### 4. Multiple `<style>` blocks — only the first is edited

`upsertStyleRule` in `server/style-edit.mjs` matches with a non-greedy, non-global regex:

```js
const styleRe = /(<style\b[^>]*>)([\s\S]*?)<\/style>/i;
const sm = source.match(styleRe);   // first match only
```

Astro components commonly have both a `<style is:global>` block (for third-party overrides) **and** a scoped `<style>` block. When `is:global` appears first, rules get injected into the global block instead of the scoped one — a subtle, hard-to-notice regression. The existing `is:global` test only covers the single-block case and won't catch this.

**Fix**: When multiple `<style>` blocks are present, prefer the one without `is:global`/`is:inline`. If none exists without those attributes, append a fresh scoped `<style>` rather than modifying the global one.

---

## Minor

### 5. Wrong refusal reason for bad input in `resolveStyle`

```js
return refuse("no-match", "edit-style value must be { property, value }");
```

`"no-match"` conventionally signals that the target element couldn't be located in source. Using it for a malformed input means a caller can't distinguish "element not found" from "the arguments were invalid." Consider `"invalid-input"` or `"invalid-value"` here.

---

### 6. Dynamic imports inside test body (`tests/apply-edit-dry-run.test.ts`)

```js
const { existsSync } = await import("node:fs");
const { join: pathJoin } = await import("node:path");
```

Both modules are already top-level imports in this file. `existsSync` can join the existing `node:fs` destructure; `pathJoin` is just the already-imported `join`.

---

davidwkeith:
Addressed all six in `68f39ac`.

**1. `windowAround` whole-file preview** — refactored to diff-based: it now finds the common prefix/suffix between `source` and `next` and clips both to a ±200 window around the changed span (signature is now `windowAround(source, next, pad)`, dropping the unused `replacement` arg). Bounded for every op, including `edit-style`'s whole-file range. Added a ~300-line-component dry-run test asserting `before.length` is far smaller than the file.

**2. `value.value` validation** — `resolveStyle` guard now also rejects `value.value === undefined` (empty string `""` still allowed — valid intent). Returns `invalid-input`. Test added for `{ property: "color" }` with no value → refused, file unchanged.

**3. `template/package.json`** — bumped to `1.1.0` (third sync file per CLAUDE.md). Note: `release.yml` only verifies `package.json` + `plugin.json` against the tag, so it wouldn't have failed on tag — but the three-file rule is explicit, so it's synced.

**4. Multiple `<style>` blocks** — `upsertStyleRule` now scans all blocks and edits the first one without `is:global`/`is:inline`; if none exists it appends a fresh scoped block rather than touching the global/inline one. This also corrected the earlier (wrong) single-`is:global` behavior — that test was replaced to expect a new scoped block. Added tests for global-first-then-scoped and global-only.

**5. Refusal reason** — malformed `edit-style` input now refuses with `invalid-input` (added to `REASON_PRIORITY`), distinct from `no-match` (element not located).

**6. Dynamic imports** — replaced with the existing top-level `node:fs`/`node:path` imports.

25/25 targeted tests pass; full suite green except the two pre-existing unrelated failures (`optimize-images-packaging`/sharp, `update.test.ts` timeout).

PR #368, opened 2026-06-21
fix(server): lazy-load sharp so a missing native dep can't crash boot (#361)

### LINKED ISSUE #361: Server crashes on load when sharp is absent — lazy-import sharp in optimize-images.mjs
## Problem

The MCP server fails to start entirely when `sharp` isn't installed, even though `sharp` is only needed by the image-optimization tool. The import is eager and on the core startup path:

- `server/apply-edit-dispatcher.mjs:34` → `import { optimizeImage } from "./optimize-images.mjs"`
- `server/optimize-images.mjs:3` → `import sharp from "sharp"` (top-level, static)

`apply-edit-dispatcher` is loaded at boot, so a missing/partial `node_modules` (no `sharp`) crashes the whole server before it ever listens:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'sharp' imported from .../server/optimize-images.mjs
    ...
Node.js v22.x
```

This is what surfaced in the app repo as a misleading "MCP e2e cold-start timing flake" (Anglesite-app#221 / PR Anglesite-app#231): the server was dead, not slow, but the client just saw a connect timeout. CI installs deps fresh so it's masked there; it bites stale/partial local checkouts.

## Proposed fix

Make `optimize-images.mjs` **lazy-import** `sharp` — a dynamic `import("sharp")` inside the handler (or a guarded loader) instead of a top-level static import. Then:

- The server boots normally without `sharp`.
- `apply_edit` and every other tool work unaffected.
- Only the image-optimization tool fails *at call time*, with a clear "image optimization requires sharp (npm install)" message instead of taking down the whole server.

## Acceptance

- [ ] `node server/index.mjs` starts and serves `tools/list` with `sharp` absent from `node_modules`.
- [ ] The image-optimization tool returns a clear, actionable error (not a process crash) when `sharp` is missing.
- [ ] With `sharp` present, image optimization behaves exactly as today.

## Notes

Worth auditing other server modules for the same eager-native-import pattern (e.g. `satori`) while here, so no single optional native dep can prevent server startup.

davidwkeith:
Closes #361.

## Problem

`server/optimize-images.mjs` sits on the MCP server's `apply_edit` boot path:

```
index.mjs → index-tools.mjs → apply-edit-dispatcher.mjs → optimize-images.mjs
```

Its top-level `import sharp from "sharp"` meant a missing/partial `sharp` (stale or incomplete `node_modules`) threw `ERR_MODULE_NOT_FOUND` during module-graph load — **crashing the whole server before it ever listened**. CI installs deps fresh so it's masked there; it bites stale local checkouts. Downstream this looked like a "MCP cold-start timing flake" (Anglesite-app#221 / #231) — the server was dead, not slow.

## Fix

Resolve `sharp` lazily through a cached `loadSharp()` helper called inside `optimizeImage`:

- The server boots normally without `sharp`.
- `apply_edit` and every other tool work unaffected.
- Only the image-optimization tool fails — **at call time**, with an actionable `"image optimization requires the 'sharp' package — run npm install"` message instead of a process crash.
- Behavior is identical when `sharp` is present.

Mirrored byte-identically into `template/server/optimize-images.mjs` (the packaging guard enforces this). Audited the rest of `server/` for the same eager-native-import pattern (the issue flagged `satori`) — **`optimize-images.mjs` was the only one**.

## Acceptance criteria

- [x] `node server/index.mjs` starts and serves with `sharp` absent
- [x] The image-optimization tool returns a clear, actionable error (not a crash) when `sharp` is missing
- [x] With `sharp` present, image optimization behaves exactly as today
- [x] Audited other server modules for the same pattern

## Tests

New `test/optimize-images-sharp-optional.test.js` simulates `sharp`'s absence with an ESM resolution hook (`test/fixtures/sharp-absent/`) run in a subprocess — faithful even though `sharp` is installed in CI. It asserts:

1. the `apply_edit` boot module graph loads when `sharp` is absent, and
2. optimization fails at call time with a clear `sharp` error.

Full suite: **2646 passed**. The one failing test (`optimize-images-packaging.test.js` scaffold e2e) is pre-existing — it fails identically on `main` because the scaffolded sandbox isn't `npm install`ed, so `sharp` isn't resolvable there. Not a regression from this change.

### DIFF
--- a/server/optimize-images.mjs
+++ b/server/optimize-images.mjs
@@ -8,20 +8,28 @@ import { join, basename, extname } from "node:path";
  * path. Loading it on first use keeps the server alive and fails only the
  * image-optimization tool — at call time, with an actionable message (#361).
  *
+ * Caches the import Promise (not the resolved value) so concurrent callers
+ * share one load, and resets on failure so a retry can succeed after a later
+ * `npm install` without restarting the server. The original failure is kept on
+ * `.cause` — a load can fail for reasons other than absence (ABI mismatch,
+ * wrong Node version, corrupted install), and discarding it would be misleading.
+ *
  * @returns {Promise<import("sharp").default>}
  */
-let _sharp;
-async function loadSharp() {
-  if (!_sharp) {
-    try {
-      _sharp = (await import("sharp")).default;
-    } catch {
-      throw new Error(
-        "image optimization requires the 'sharp' package — run `npm install` (or `npm install sharp`)",
-      );
-    }
+let _sharpPromise;
+function loadSharp() {
+  if (!_sharpPromise) {
+    _sharpPromise = import("sharp")
+      .then((m) => m.default)
+      .catch((cause) => {
+        _sharpPromise = undefined;
+        throw new Error(
+          "image optimization requires the 'sharp' package — run `npm install` (or `npm install sharp`)",
+          { cause },
+        );
+      });
   }
-  return _sharp;
+  return _sharpPromise;
 }
 
 /**
--- a/server/optimize-images.mjs
+++ b/server/optimize-images.mjs
@@ -1,6 +1,28 @@
 import { mkdirSync, copyFileSync, existsSync } from "node:fs";
 import { join, basename, extname } from "node:path";
-import sharp from "sharp";
+
+/**
+ * Lazily resolve `sharp`. It's an optional native dependency: importing it at
+ * module top level would crash the whole MCP server at boot when it's absent
+ * (stale/partial node_modules), because this module sits on the apply_edit boot
+ * path. Loading it on first use keeps the server alive and fails only the
+ * image-optimization tool — at call time, with an actionable message (#361).
+ *
+ * @returns {Promise<import("sharp").default>}
+ */
+let _sharp;
+async function loadSharp() {
+  if (!_sharp) {
+    try {
+      _sharp = (await import("sharp")).default;
+    } catch {
+      throw new Error(
+        "image optimization requires the 'sharp' package — run `npm install` (or `npm install sharp`)",
+      );
+    }
+  }
+  return _sharp;
+}
 
 /**
  * Optimize a single image: write a primary WebP plus responsive variants,
@@ -18,6 +40,7 @@ import sharp from "sharp";
  * }>}
  */
 export async function optimizeImage(inputFile, options) {
+  const sharp = await loadSharp();
   const widths = options.widths ?? [480, 768, 1024, 1920];
   const outputDir = options.outputDir;
   if (!existsSync(outputDir)) mkdirSync(outputDir, { recursive: true });
--- a/test/fixtures/sharp-absent/hooks.mjs
+++ b/test/fixtures/sharp-absent/hooks.mjs
@@ -3,9 +3,9 @@
 // Used via `node --import ./register.mjs <probe>` in optimize-images-sharp-optional.test.js.
 export async function resolve(specifier, context, nextResolve) {
   if (specifier === "sharp") {
-    const err = new Error(
-      "Cannot find package 'sharp' imported from optimize-images.mjs (simulated absent)",
-    );
+    // The hook intercepts any `import "sharp"` in the process, so the message
+    // stays attribution-free rather than naming a specific importer.
+    const err = new Error("Cannot find package 'sharp' (simulated absent for testing #361)");
     err.code = "ERR_MODULE_NOT_FOUND";
     throw err;
   }
--- a/test/fixtures/sharp-absent/hooks.mjs
+++ b/test/fixtures/sharp-absent/hooks.mjs
@@ -1,0 +1,13 @@
+// ESM resolution hook that simulates `sharp` being absent from node_modules,
+// reproducing the ERR_MODULE_NOT_FOUND a stale/partial checkout hits (#361).
+// Used via `node --import ./register.mjs <probe>` in optimize-images-sharp-optional.test.js.
+export async function resolve(specifier, context, nextResolve) {
+  if (specifier === "sharp") {
+    const err = new Error(
+      "Cannot find package 'sharp' imported from optimize-images.mjs (simulated absent)",
+    );
+    err.code = "ERR_MODULE_NOT_FOUND";
+    throw err;
+  }
+  return nextResolve(specifier, context);
+}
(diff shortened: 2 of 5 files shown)


server/optimize-images.mjs
```diff
@@ -1,6 +1,28 @@
...
+ */
+let _sharp;
+async function loadSharp() {
+  if (!_sharp) {
+    try {
+      _sharp = (await import("sharp")).default;
+    } catch {
+      throw new Error(
+        "image optimization requires the 'sharp' package — run `npm install` (or `npm install sharp`)",
+      );
```
davidwkeith: The original error is silently discarded here. When `sharp` fails to load for a reason *other* than being absent — mismatched native binary, ABI incompatibility, wrong Node.js version, a corrupted install — the caller sees only `"run npm install"`, which is actively misleading and the real cause is gone.

```js
} catch (cause) {
  throw new Error(
    "image optimization requires the 'sharp' package — run `npm install` (or `npm install sharp`)",
    { cause },
  );
}
```

Adding `{ cause }` keeps the original error in `.cause` so it shows up in stack traces and via `err.cause` if callers inspect it.

---

server/optimize-images.mjs
```diff
@@ -1,6 +1,28 @@
...
+ * module top level would crash the whole MCP server at boot when it's absent
+ * (stale/partial node_modules), because this module sits on the apply_edit boot
+ * path. Loading it on first use keeps the server alive and fails only the
+ * image-optimization tool — at call time, with an actionable message (#361).
+ *
+ * @returns {Promise<import("sharp").default>}
+ */
+let _sharp;
+async function loadSharp() {
+  if (!_sharp) {
```
davidwkeith: Minor: if `optimizeImage` is called concurrently before the first `loadSharp()` settles, multiple callers pass `!_sharp` and each fires `import("sharp")` in parallel. Harmless in practice (the Node.js module loader deduplicates dynamic imports under the hood), but caching the Promise instead of the resolved value is the idiomatic pattern and makes the intent clearer:

```js
let _sharpPromise;
async function loadSharp() {
  if (!_sharpPromise) {
    _sharpPromise = import("sharp")
      .then((m) => m.default)
      .catch((cause) => {
        _sharpPromise = undefined; // reset so a retry is possible after npm install
        throw new Error(
          "image optimization requires the 'sharp' package — run `npm install`",
          { cause },
        );
      });
  }
  return _sharpPromise;
}
```

Not blocking, but worth considering — especially since the reset-on-failure path also means a subsequent call after a successful `npm install` would work without a server restart.

---

test/optimize-images-sharp-optional.test.js
```diff
@@ -0,0 +1,55 @@
...
+const REGISTER = join(__dirname, "fixtures", "sharp-absent", "register.mjs");
+const DISPATCHER = join(REPO_ROOT, "server", "apply-edit-dispatcher.mjs");
+const OPTIMIZE = join(REPO_ROOT, "server", "optimize-images.mjs");
+
+/** Run a snippet in a subprocess with `sharp` made unresolvable. Returns stdout. */
+function runWithoutSharp(snippet) {
+  return execFileSync(
+    process.execPath,
+    ["--import", REGISTER, "--input-type=module", "-e", snippet],
+    { cwd: REPO_ROOT, encoding: "utf-8", stdio: ["ignore", "pipe", "pipe"] },
```
davidwkeith: `stdio: ["ignore", "pipe", "pipe"]` pipes stderr but `execFileSync` only returns stdout. If the subprocess crashes in an unexpected way (not caught inside the snippet), `execFileSync` throws a `SpawnError` and Vitest reports only the generic `"Command failed"` message — the actual stderr with the stack trace is buried in `err.stderr` and easy to miss in CI logs.

Either inherit stderr so it streams to the test runner output:

```js
{ cwd: REPO_ROOT, encoding: "utf-8", stdio: ["ignore", "pipe", "inherit"] }
```

Or catch and re-throw with stderr attached:

```js
try {
  return execFileSync(process.execPath, [...], { cwd: REPO_ROOT, encoding: "utf-8" });
} catch (e) {
  throw new Error(`subprocess failed:\n${e.stderr}\n${e.stdout}`, { cause: e });
}
```

Either makes CI failures in these tests much easier to diagnose.

---

test/fixtures/sharp-absent/hooks.mjs
```diff
@@ -0,0 +1,13 @@
+// ESM resolution hook that simulates `sharp` being absent from node_modules,
+// reproducing the ERR_MODULE_NOT_FOUND a stale/partial checkout hits (#361).
+// Used via `node --import ./register.mjs <probe>` in optimize-images-sharp-optional.test.js.
+export async function resolve(specifier, context, nextResolve) {
+  if (specifier === "sharp") {
+    const err = new Error(
+      "Cannot find package 'sharp' imported from optimize-images.mjs (simulated absent)",
```
davidwkeith: The error message hardcodes `"imported from optimize-images.mjs"`. The hook intercepts *any* module that tries to `import "sharp"` in the process, not just `optimize-images.mjs`. If a future test probe imports another module that also tries to import sharp, the attribution will be wrong. Removing the `"imported from …"` clause is simpler and still produces a faithful `ERR_MODULE_NOT_FOUND`:

```js
const err = new Error(
  "Cannot find package 'sharp' (simulated absent for testing #361)",
);
```

---

davidwkeith (commented):
The lazy-load approach is the right fix — moving `sharp` out of the module graph so a missing native dep can't crash the server at boot is exactly correct, and the subprocess-based test strategy is solid. Four things to address before merging:

**Blocker:** The `catch {}` in `loadSharp()` silently drops the original error. When `sharp` is *present* but fails (ABI mismatch, wrong Node.js version, corrupted install), the message `"run npm install"` is actively misleading and the real cause is gone. Fix: `catch (cause) { throw new Error("...", { cause }); }`.

**Suggestions:**
- Cache the Promise rather than the resolved value in `loadSharp()` — eliminates the redundant concurrent-import race and, as a bonus, enables a clean reset-on-failure path so callers can recover after `npm install` without a server restart (see inline comment).
- `execFileSync` with `stdio: ["ignore", "pipe", "pipe"]` buries subprocess stderr in CI; switch to `"inherit"` for stderr or re-throw with it appended so failures in these tests are diagnosable.

**Nit:** The fixture `hooks.mjs` error message hardcodes `"imported from optimize-images.mjs"` but the hook intercepts any importer — minor but could mislead when debugging future test probes.

---

Issue #363, opened 2026-06-18
1.1.0 inbound IndieWeb broken: @dwk beta packages don't load, template/package API mismatch, stored XSS in webmention render


davidwkeith:
Found while wiring up inbound IndieWeb during a `beta.6 → 1.1.0` re-platform (adopting `worker/site-entry.js` + the `@dwk/*` packages per `skills/indieweb/SKILL.md`). Three separate, blocking problems — the scaffold only "builds" because the IndieWeb code is inert (no D1 bindings) and mocked in tests.

**Versions:** `anglesite@1.1.0` template, `@dwk/indieauth`, `@dwk/micropub`, `@dwk/webmention` all `0.1.0-beta.2`. Node 22, Astro 6.3.1.

---

### 1. The published `@dwk/*` beta packages don't load (Node ESM)

Each package's `dist/index.js` re-exports from extensionless relative paths that Node ESM can't resolve:

```
$ node -e "import('@dwk/webmention').then(m=>console.log(Object.keys(m)))"
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '.../@dwk/webmention/dist/inbox'
  imported from .../@dwk/webmention/dist/index.js
```

Same for `@dwk/indieauth` (`.../dist/handler`) and `@dwk/micropub` (`.../dist/handler`). The dist needs explicit `.js` extensions on relative re-exports (or `"moduleResolution": "bundler"` won't help consumers — Node ESM requires extensions). esbuild/wrangler may paper over this when bundling, but the packages are unusable in plain Node (and untestable in a Vitest/Node worker test without mocking the whole module).

### 2. Template ↔ package API mismatch in `worker/site-entry.js`

The template imports a `createHandler` alias that the packages don't export:

```js
// template/worker/site-entry.js:55-57
import { createHandler as createIndieAuth } from "@dwk/indieauth";
import { createHandler as createMicropub } from "@dwk/micropub";
import { createHandler as createWebmention } from "@dwk/webmention";
```

[...]

PR #369, opened 2026-06-21
fix(indieweb): block dangerous URL schemes in webmention render (XSS)

davidwkeith:
## Problem (HIGH — stored XSS)

`renderMention()` in `template/worker/site-entry.js` interpolates externally-sourced webmention fields (`author_url`, `author_photo`, `url`) into `href`/`src` with only `escapeHtml()`. Escaping blocks tag injection but **not dangerous URL schemes** — a verified webmention whose `author_url`/`url` is `javascript:…` (or `data:…`) rendered an executable link/`src`. Webmention data originates from arbitrary external sites, so this is stored XSS on any page with mentions once Webmention is enabled.

## Fix

- `safeUrl(value)` — parse and allow through only `http:`/`https:`; otherwise drop the attribute. The author name still renders, just not as a link.
- Add `rel="nofollow ugc noopener"` to mention anchors (external, user-generated content shouldn't pass link equity or a window handle).

## Tests (TDD)

`tests/indieweb-render-safety.test.ts` — written failing first, then fixed:
- `javascript:` / `JavaScript:` / ` javascript:` / `data:` / `vbscript:` are dropped from `href` and `src`
- `http`/`https` pass through, with `nofollow`/`ugc`/`noopener` present
- text fields remain HTML-escaped

26/26 pass (render-safety + existing dispatch suite). Scoped to this fix only; does not touch the in-flight #361 sharp work or the broader #363 API-alignment effort.

Addresses problem 3 of #363.

### DIFF
--- a/template/worker/site-entry.js
+++ b/template/worker/site-entry.js
@@ -304,21 +304,41 @@ class WebmentionInjector {
   }
 }
 
-function renderMention(m) {
+export function renderMention(m) {
   const name = escapeHtml(m.author_name || m.author_url || "Someone");
-  const photo = m.author_photo
-    ? `<img class="u-photo" src="${escapeHtml(m.author_photo)}" alt="" width="48" height="48" />`
+  // Webmention fields come from arbitrary external sites. escapeHtml() blocks
+  // tag injection but NOT dangerous URL schemes — a `javascript:`/`data:` value
+  // in an href/src is executable. Only http/https URLs may reach an attribute.
+  const photoUrl = safeUrl(m.author_photo);
+  const authorUrl = safeUrl(m.author_url);
+  const url = safeUrl(m.url);
+  const photo = photoUrl
+    ? `<img class="u-photo" src="${escapeHtml(photoUrl)}" alt="" width="48" height="48" />`
     : "";
-  const author = m.author_url
-    ? `<a class="p-author h-card u-url" href="${escapeHtml(m.author_url)}">${name}</a>`
+  const author = authorUrl
+    ? `<a class="p-author h-card u-url" rel="nofollow ugc noopener" href="${escapeHtml(authorUrl)}">${name}</a>`
     : `<span class="p-author">${name}</span>`;
   const content = m.content
     ? `<div class="p-content">${escapeHtml(m.content)}</div>`
     : "";
-  const permalink = m.url
-    ? `<a class="u-url" href="${escapeHtml(m.url)}">permalink</a>`
+  const permalink = url
+    ? `<a class="u-url" rel="nofollow ugc noopener" href="${escapeHtml(url)}">permalink</a>`
     : "";
   return `<li class="h-cite">${photo}${author}${content}${permalink}</li>`;
+}
+
+// Return the URL only if it parses to an http(s) URL, else null. Blocks
+// javascript:/data:/vbscript:/mailto: and any other scheme from reaching an
+// href/src attribute. Relative URLs are rejected (no base to resolve against
+// here); webmention author/permalink URLs are absolute in practice.
+export function safeUrl(value) {
+  if (!value) return null;
+  try {
+    const u = new URL(String(value).trim());
+    return u.protocol === "http:" || u.protocol === "https:" ? u.href : null;
+  } catch {
+    return null;
+  }
 }
 
 function escapeHtml(value) {
--- a/template/worker/site-entry.js
+++ b/template/worker/site-entry.js
@@ -316,13 +316,13 @@ export function renderMention(m) {
     ? `<img class="u-photo" src="${escapeHtml(photoUrl)}" alt="" width="48" height="48" />`
     : "";
   const author = authorUrl
-    ? `<a class="p-author h-card u-url" rel="nofollow ugc noopener" href="${escapeHtml(authorUrl)}">${name}</a>`
+    ? `<a class="p-author h-card u-url" rel="nofollow ugc noopener noreferrer" href="${escapeHtml(authorUrl)}">${name}</a>`
     : `<span class="p-author">${name}</span>`;
   const content = m.content
     ? `<div class="p-content">${escapeHtml(m.content)}</div>`
     : "";
   const permalink = url
-    ? `<a class="u-url" rel="nofollow ugc noopener" href="${escapeHtml(url)}">permalink</a>`
+    ? `<a class="u-url" rel="nofollow ugc noopener noreferrer" href="${escapeHtml(url)}">permalink</a>`
     : "";
   return `<li class="h-cite">${photo}${author}${content}${permalink}</li>`;
 }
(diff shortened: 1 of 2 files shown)


template/worker/site-entry.js
```diff
@@ -304,23 +304,43 @@ class WebmentionInjector {
...
+  const photoUrl = safeUrl(m.author_photo);
+  const authorUrl = safeUrl(m.author_url);
+  const url = safeUrl(m.url);
+  const photo = photoUrl
+    ? `<img class="u-photo" src="${escapeHtml(photoUrl)}" alt="" width="48" height="48" />`
     : "";
-  const author = m.author_url
-    ? `<a class="p-author h-card u-url" href="${escapeHtml(m.author_url)}">${name}</a>`
+  const author = authorUrl
+    ? `<a class="p-author h-card u-url" rel="nofollow ugc noopener" href="${escapeHtml(authorUrl)}">${name}</a>`
```
davidwkeith: Missing `noreferrer` here and on the permalink anchor below (line 325).

The canonical hardening set for untrusted external links is `nofollow ugc noopener noreferrer`. `noreferrer` implies `noopener` and additionally suppresses the `Referer` header — you don't want a hostile site to receive a signal that it appeared as a webmention on this domain.

```js
rel="nofollow ugc noopener noreferrer"
```

---

template/worker/site-entry.js
```diff
@@ -304,23 +304,43 @@ class WebmentionInjector {
...
     : `<span class="p-author">${name}</span>`;
   const content = m.content
     ? `<div class="p-content">${escapeHtml(m.content)}</div>`
     : "";
-  const permalink = m.url
-    ? `<a class="u-url" href="${escapeHtml(m.url)}">permalink</a>`
+  const permalink = url
+    ? `<a class="u-url" rel="nofollow ugc noopener" href="${escapeHtml(url)}">permalink</a>`
```
davidwkeith: Same — `noreferrer` missing. Both anchors should use `rel="nofollow ugc noopener noreferrer"`.

---

tests/indieweb-render-safety.test.ts
```diff
@@ -0,0 +1,72 @@
...
+import { describe, it, expect } from "vitest";
+import { renderMention } from "../template/worker/site-entry.js";
+
+const HOSTILE = [
+  "javascript:alert(1)",
+  "JavaScript:alert(1)",
+  " javascript:alert(1)",
+  "data:text/html,<script>alert(1)</script>",
+  "vbscript:msgbox(1)",
+  "javascript:alert(1)",
```
davidwkeith: This last entry is a byte-for-byte duplicate of `HOSTILE[0]` — `"javascript:alert(1)"` appears twice. This slot was presumably intended for a Unicode lookalike such as `"ｊavascript:alert(1)"` (U+FF4A FULLWIDTH LATIN SMALL LETTER J). The WHATWG URL parser rejects that too (non-ASCII is not a valid scheme start), so `safeUrl` blocks it via the `catch` — but it's worth exercising explicitly since Unicode-confusable schemes are a documented bypass class.

```ts
"ｊavascript:alert(1)",   // U+FF4A fullwidth 'j' — rejected by URL scheme parsing
```

---

tests/indieweb-render-safety.test.ts
```diff
@@ -0,0 +1,72 @@
...
+];
+
+describe("renderMention URL-scheme safety", () => {
+  it("never emits a javascript:/data:/vbscript: href for author_url", () => {
+    for (const url of HOSTILE) {
+      const html = renderMention({ author_name: "X", author_url: url });
+      expect(html, url).not.toMatch(/href="[^"]*(javascript|data|vbscript):/i);
+      // The name still renders, just not as a link.
```
davidwkeith, line 31: All hostile-URL tests supply an explicit `author_name: "X"`, so the name-fallback path isn't exercised. When `author_name` is absent and `author_url` is a hostile scheme, the raw hostile URL becomes the visible text of the fallback `<span>` — safe (it's HTML-escaped), but it's undocumented behaviour that a future reader could "fix" in a way that puts the value back into an attribute. Worth a dedicated test:

```ts
it("degrades gracefully when only author_url is hostile (name falls back to escaped text)", () => {
  const html = renderMention({ author_url: "javascript:alert(1)" });
  expect(html).not.toContain("href=");
  expect(html).not.toContain("<a ");
  // Raw URL is visible as escaped text — safe, not executable
  expect(html).toContain("javascript:alert(1)");
});
```

---

davidwkeith (commented):
The core fix is correct and well-structured — `safeUrl()` with a strict http/https allowlist is the right approach, the positive-allowlist pattern is robust against future scheme additions, and the TDD suite covers the main attack classes. Three things to address:

**Must fix**
- Both new anchors use `rel="nofollow ugc noopener"` but are missing `noreferrer`. For UGC links you want to suppress the `Referer` header as well — you don't want a hostile site to receive a signal that it appeared as a webmention. The full canonical set is `nofollow ugc noopener noreferrer`. (`noreferrer` also implies `noopener`, so the two together are not redundant.)

**Test quality**
- `HOSTILE[5]` (line 19) is a byte-for-byte duplicate of `HOSTILE[0]`. Replace it with a Unicode-confusable lookalike (`ｊavascript:` with U+FF4A) — the most common bypass class that slot typically guards against.
- No test covers the `author_name`-absent path where a hostile `author_url` falls back to visible escaped text. Safe today, but explicit coverage prevents someone from inadvertently putting the value back into an attribute when refactoring the fallback logic.

---

davidwkeith (commented):
The fix is correct and the approach (allow-list on scheme, not deny-list) is exactly right. `safeUrl()` using `URL.protocol` is solid — it handles scheme-casing and leading whitespace via `.trim()`, the catch covers malformed values, and `u.href` returns the normalized form. `rel="nofollow ugc noopener"` on all mention anchors is the right default.

One test quality issue: `HOSTILE` has `"javascript:alert(1)"` at index 0 **and** again at index 5 — the last entry is a duplicate of the first. A missing test case would be `"mailto:me@example.com"` or `"file:///etc/passwd"`. (This is fixed in the stacked #370, which changes index 5 to `"mailto:me@example.com"` — just noting so it isn't re-introduced if this PR ever needs a rebase.)

---

template/worker/site-entry.js
davidwkeith: Fixed in 71c7367 — both the author and permalink anchors now use `rel="nofollow ugc noopener noreferrer"`.

template/worker/site-entry.js
davidwkeith: Done — permalink anchor updated to `nofollow ugc noopener noreferrer` in 71c7367.

tests/indieweb-render-safety.test.ts
davidwkeith: Fixed in 71c7367 — HOSTILE[5] dup replaced with mailto:, file://, and the U+FF4A fullwidth-'j' confusable (rejected by the WHATWG parser, so safeUrl drops it via catch).

tests/indieweb-render-safety.test.ts
davidwkeith: Added a dedicated test in 71c7367: author_name absent + hostile author_url renders only as escaped text in the fallback span, asserting it never reaches href/src.

PR #370, opened 2026-06-21
fix(indieweb): realign Webmention wiring to the real @dwk/webmention API

davidwkeith:
Stacked on #369 (the XSS fix it builds on — `safeUrl` is reused).

## Problem

`template/worker/site-entry.js` was written against an **imagined** `@dwk/webmention` API. The shipped `0.1.0-beta.3` differs in every load-bearing detail, so the inbound-webmention path cannot work:

| Assumed (template) | Real (`@dwk/webmention@0.1.0-beta.3`) |
|---|---|
| `createHandler()` (no config) | `createWebmention({ baseUrl })` |
| handler has `.queue()` / `.scheduled()` | queue is a **separate** `createWebmentionQueueConsumer(config)`; no `scheduled` arm |
| reads `env.WEBMENTION_DB` | reads `env.WEBMENTION_INBOX` (D1) + `env.WEBMENTION_QUEUE` |
| inbox rows carry `author_*, content, url, type, status` | inbox stores only `{ source, target, verified_at }` |

The edge-render selected columns that don't exist → it threw → caught → "no mentions" → **mentions never rendered**.

## Changes

- **Imports/construction:** `createWebmention` + the separate `createWebmentionQueueConsumer` + `createD1Inbox`, built per-`env` (memoized) with `baseUrl` from a new **`SITE_URL`** var (works off-request, where the queue consumer has no request URL).
- **Binding rename** `WEBMENTION_DB` → `WEBMENTION_INBOX` across worker, `wrangler.jsonc`, skill, docs, and page comments.
- **`queue()`** runs the queue consumer; **`scheduled()`** drops the nonexistent webmention arm and keeps the Micropub bridge sync.
- **Display:** the inbox only stores `{ source, target, verified_at }`, so the edge-render is now a minimal **"Mentioned by" source-link list** read via `createD1Inbox().list()`. `safeUrl` (from #369) still guards the `source` href; `rel="nofollow ugc noopener"`.

## Tests (TDD)

Stub rewritten to the real API; dispatch + wiring + render-safety updated. **41/41** IndieWeb tests pass; full suite green except the pre-existing #361 optimize-images failure (separate branch).

Design spec: `docs/superpowers/specs/2026-06-21-webmention-realignment-design.md`.

## Scope

Webmention slice of problem 2 of #363. **IndieAuth (passkey `approveAuthorization`) and Micropub realignment are tracked separately** — IndieAuth needs a real design (it's the next spec).

> Depends on the `@dwk/workers` ESM fix (davidwkeith/workers#165) shipping as `beta.3` for the packages to load in Node; esbuild/wrangler already bundle past it at deploy.


davidwkeith (commented):
The realignment is thorough — binding rename propagated everywhere, the API mismatch is fully fixed, and the memoized-per-env WeakMap pattern is the right way to build the factories at request time without re-constructing them on every call. Three things should be fixed before merge.

---

### 1. Dead code in `renderMention()` — the inner `try/catch` can never fire

```js
export function renderMention(m) {
  const href = safeUrl(m.source);  // returns u.href — already a valid URL
  if (!href) return "";
  let host;
  try {
    host = new URL(href).host;     // ← can never throw; href is guaranteed valid
  } catch {
    host = href;                   // ← unreachable
  }
  …
}
```

`safeUrl()` calls `new URL(…)` and returns `u.href` only on success. `href` is therefore a well-formed URL string; the subsequent `new URL(href).host` cannot throw. Remove the `try/catch`:

```js
export function renderMention(m) {
  const href = safeUrl(m.source);
  if (!href) return "";
  const host = new URL(href).host;
  return (
    `<li class="h-cite">` +
    `<a class="u-url" rel="nofollow ugc noopener" href="${escapeHtml(href)}">${escapeHtml(host)}</a>` +
    `</li>`
  );
}
```

---

### 2. Empty `SITE_URL` silently breaks the webmention receiver

`wrangler.jsonc` now ships `"SITE_URL": ""` as the default. If a user provisions `WEBMENTION_INBOX` (D1 database ID filled in) but doesn't complete the `SITE_URL` step, `webmentionFor(env)` is called with `baseUrl: ""`. `@dwk/webmention`'s receiver uses `baseUrl` for target-URL validation — an empty `baseUrl` means it will reject every inbound mention, silently, with no error surfaced to the operator.

The worker should guard before constructing the factories:

```js
function webmentionFor(env) {
  let bundle = webmentionByEnv.get(env);
  if (!bundle) {
    const baseUrl = env.SITE_URL;
    if (!baseUrl) {
      console.error(
        "SITE_URL is not set — @dwk/webmention cannot validate targets. " +
        "Run /anglesite:indieweb to configure it.",
      );
    }
    bundle = {
      handler: createWebmention({ baseUrl }),
      consumer: createWebmentionQueueConsumer({ baseUrl }),
      inbox: env.WEBMENTION_INBOX ? createD1Inbox(env.WEBMENTION_INBOX) : null,
    };
    webmentionByEnv.set(env, bundle);
  }
  return bundle;
}
```

A test covering "WEBMENTION_INBOX bound but SITE_URL empty → `console.error` called" would prevent this from regressing silently.

---

### 3. Spec file in undocumented directory

`docs/superpowers/specs/2026-06-21-webmention-realignment-design.md` is committed under `docs/superpowers/specs/`, a path that doesn't exist in the project's documented directory structure (CLAUDE.md lists `docs/decisions/` as the home for architectural decisions). The ADR-0020 inline update in this PR already captures the key decisions there. The spec document could either live in `docs/decisions/` as ADR-0022, or be dropped — the PR description and the ADR update together are sufficient record.

---

davidwkeith:
All three addressed in the rebased branch (`daaf6fe`):

1. **Dead `try/catch` in `renderMention()`** — removed. `safeUrl()` returns a normalized valid URL string, so the `new URL(href).host` reparse can't throw; it's now a plain `const host`.
2. **Empty `SITE_URL` silently breaks the receiver** — `webmentionFor(env)` now `console.error`s once (memoized per-env) when `WEBMENTION_INBOX` is bound but `SITE_URL` is empty, naming `/anglesite:indieweb` as the fix. Covered by two new tests in `indieweb-dispatch.test.ts` (warns when empty / silent when set).
3. **Spec file location** — dropped. The ADR-0020 inline update in this PR is the record, per your note.

Also carried the `noreferrer` hardening from #369 onto the (now single) mention anchor: `rel="nofollow ugc noopener noreferrer"`.

Rebased onto `main` after #369 merged; the diff is now just the Webmention realignment. IndieWeb suite 43/43 green.

PR #371, opened 2026-06-21
feat(indieweb): passkey-backed IndieAuth — spec, plan & implementation (#363)

davidwkeith:
Implements the **IndieAuth slice of #363, problem 2** — `@dwk/indieauth`'s required `approveAuthorization` hook wired to a passkey (`@dwk/webauthn`) owner-authentication + consent flow, with printable single-use backup codes. Spec + plan + implementation, built test-first.

## What's here
- **Owner-auth core** (`template/worker/owner-auth.js`): signed owner-session cookie, request-bound consent token (HMAC over `client_id|redirect_uri|scope|exp`), single-use backup codes (SHA-256 in `OWNER_AUTH_DB`).
- **Consent flow** (`site-entry.js` + `auth-pages.js`): `approveAuthorization` returns the consent page until a valid session **and** consent token exist, then `AuthorizationApproval { me: SITE_URL }`. `POST /auth/consent` redeems a passkey assertion or backup code → session + consent token → 302 back to `/auth`.
- **WebAuthn ceremonies**: `WEBAUTHN` Durable Object (first DO in the template, with `migrations`); `/auth/webauthn/*` mounted; `/register/*` gated owner-only (one-time `INDIEWEB_REG_TOKEN` or an existing session); `GET /auth/register` page.
- **Provisioning**: `OWNER_AUTH_DB` binding, SKILL Step 5c (secrets, one-time link, backup codes, recovery/reset), deploy-scan rejects committed `INDIEAUTH_SESSION_KEY`/`INDIEWEB_REG_TOKEN` (both the bash hook and the TS `runScan`), `dwk-workers.md`, **ADR-0022**.

## Security
- Consent token blocks consent-screen-skip deep links; `user_verification` required; rpId/origin domain-bound; registration fails closed (403); backup codes hashed + single-use with no unknown-code oracle.

## Tests (TDD, 9 tasks)
~40 new tests across session / consent-token / backup-codes / approveAuthorization / consent-endpoint / dispatch / registration-gating / deploy-scan. Full suite green (excluding the pre-existing local `optimize-images` sandbox artifact).

Design: `docs/superpowers/specs/2026-06-21-passkey-indieauth-design.md` · Plan: `docs/superpowers/plans/2026-06-21-passkey-indieauth.md`.

> Depends on `@dwk/* beta.3` (workers#165, merged) being published for real-Node loading; tests use aliased stubs, and wrangler/esbuild bundle past it at deploy. Micropub realignment (#2c) is separate.

### DIFF
--- a/docs/superpowers/specs/2026-06-21-passkey-indieauth-design.md
+++ b/docs/superpowers/specs/2026-06-21-passkey-indieauth-design.md
@@ -1,0 +1,255 @@
+# Passkey-backed IndieAuth (`approveAuthorization`)
+
+**Date:** 2026-06-21 · **Issue:** Anglesite/anglesite#363 (problem 2, IndieAuth slice) · **Status:** draft — awaiting review
+
+Successor to `2026-06-09-active-indieweb-design.md` and ADR-0020. Sibling of the
+Webmention realignment (shipped, #370). Micropub realignment is a separate spec
+(#2c) that depends on this one.
+
+## Problem
+
+`@dwk/indieauth` owns all IndieAuth protocol mechanics (PKCE, code/token
+issuance, metadata) but **deliberately delegates authentication and consent to
+the deployer**:
+
+```ts
+readonly approveAuthorization: ApproveAuthorization; // required config
+type ApproveAuthorization =
+  (request: AuthorizationRequest, httpRequest: Request)
+    => Promise<AuthorizationApproval | Response>;
+```
+
+The current template calls `createIndieAuth()` with no config — it cannot even
+construct, and there is no mechanism for the owner to prove they are the owner
+before a token is minted. We need to build that mechanism. Per brainstorming
+(2026-06-21) the chosen mechanism is **passkeys via `@dwk/webauthn`**.
+
+## The contract we build against
+
+`approveAuthorization` returns one of two things:
+
+- **`AuthorizationApproval { me, scopes?, profile? }`** — the library mints a code
+  and redirects to the client. Return this only when the current request carries
+  a valid owner session **and** a consent token bound to this exact request.
+- **`Response`** — the library returns it verbatim, handing us the exchange. This
+  is the escape hatch for interactive auth: we return the login/consent page.
+
+This single seam is the whole integration point. Everything else is the page,
+the session, and the credential store behind it.
+
+## What `@dwk/webauthn` gives us (and what it doesn't)
+
+`createWebAuthn({ rpId, rpName, origin })` → a `fetch` handler exposing four
+ceremony endpoints, backed by a **`WEBAUTHN` Durable Object namespace**
+(`WebAuthnObject`, exported by the package) that holds challenge state +
+credential records (strongly consistent):
+
+| Endpoint | Purpose |
+|---|---|
+| `POST /register/options` | issue creation options + challenge |
+| `POST /register/verify` | verify attestation, **store the credential** |
+| `POST /authenticate/options` | issue request options + challenge |
+| `POST /authenticate/verify` | verify assertion, advance the counter |
+
+**Critical gap:** the package is a generic relying party. Its `/register/*`
+endpoints will enrol **any** passkey that completes the ceremony — it has no
+notion of "the owner." Gating registration so only the owner can enrol the first
+credential is **our** responsibility. This is the load-bearing design decision
+(see Open decisions §1).
+
+> Caveat to record: `@dwk/webauthn` is self-described "exploratory, lowest
+> priority" — the least-mature package in the set. We are putting the owner's
+> primary identity on it. Acceptable given the owner controls the whole stack and
+> can reset auth (§ Recovery), but worth stating plainly.
+
+## Architecture
+
+Everything mounts under `/auth` in `site-entry.js`, ahead of the `ASSETS`
+fallthrough, gated on `env.AUTH_DB` (as today):
+
+```
+/auth                       → @dwk/indieauth authorization + token + metadata
+/auth/webauthn/register/*    → @dwk/webauthn (owner-only gated, see §1)
+/auth/webauthn/authenticate/* → @dwk/webauthn (open — proves possession only)
+/auth/register               → owner-only registration page (HTML)
+/auth/consent                → POST target: validate session, mint consent token
+```
+
+Per-request construction, memoized per `env` (the `webmentionFor` pattern from
+#370), because `createIndieAuth`/`createWebAuthn` need config (`baseUrl`,
+`rpId`, `origin`) that derives from `env.SITE_URL`.
+
+### The flow
+
+```
+1. Client → GET /auth?response_type=code&client_id&redirect_uri&state
+            &code_challenge&code_challenge_method=S256&scope&me
+2. @dwk/indieauth validates → approveAuthorization(request, httpRequest)
+3. our hook reads the signed __anglesite_owner session cookie:
+   a. no/expired session →
+        return Response: consent page (HTML). The page:
+          - runs a WebAuthn assertion via /auth/webauthn/authenticate/*
+            (navigator.credentials.get)
+          - on success POSTs to /auth/consent with the assertion result +
+            the original authorization params
+          - /auth/consent verifies the assertion server-side, sets the
+            __anglesite_owner session cookie, mints a per-request consent
+            token (HMAC over client_id|redirect_uri|scope|nonce|exp), and
+            302-redirects back to /auth?...original...&_consent=<token>
+   b. session present + valid _consent token for THIS request →
+        return AuthorizationApproval { me: SITE_URL, scopes, profile }
+   c. session present, no consent token →
+        return Response: consent page in "already authenticated" mode
+        (skip the passkey step, just show client+scopes, Approve → mint token)
+4. @dwk/indieauth mints the code, redirects to redirect_uri with code+state
+```
+
+The consent token is what prevents a malicious client from deep-linking
+`/auth?...&_consent=…` to skip the visible consent screen: the token is an HMAC
+the owner's browser only receives after POSTing through `/auth/consent`, bound to
+the specific request parameters and short-lived.
+
+### Session
+
+A signed cookie `__anglesite_owner`, reusing the existing membership HMAC helper
+shape in `site-entry.js` (`verifyMembershipCookie`/`crypto.subtle` HMAC-SHA-256).
+Payload `{ sub: "owner", iat, exp }`, `exp` ~15 min, `HttpOnly; Secure; SameSite=Lax`.
+A fresh passkey assertion re-issues it. Signing key: a new
+`INDIEAUTH_SESSION_KEY` secret (or reuse `INDIEAUTH_SIGNING_KEY`).
+
+### Registration
+
+The owner enrols a passkey at `/auth/register` (HTML page driving
+`/auth/webauthn/register/*`). Gated owner-only (§1). The page encourages
+enrolling **two** credentials (e.g. phone + laptop, or a hardware key) so a lost
+device isn't a lockout. Once authenticated, the same page (reached with a valid
+session) lets the owner add or remove credentials.
+
+### Backup codes
+
+A second recovery factor alongside multiple passkeys and the redeploy root. At
+setup (and regenerable from an authenticated session), the worker generates **10
+single-use backup codes**, shows them once for the owner to print/save, and
+stores only their hashes. The consent page offers "Use a backup code instead of a
+passkey"; a submitted code is hashed and matched against an unused stored hash —
+on match it establishes the owner session and the code is marked consumed. A used
+or unknown code returns the consent page unchanged (no oracle).
+
+- **Storage:** a small owned table (`owner_backup_codes`: `code_hash`,
+  `created_at`, `used_at`) in a dedicated `OWNER_AUTH_DB` D1 binding — *not* in
+  `AUTH_DB`, whose schema belongs to `@dwk/indieauth`. Single-use is enforced by
+  `used_at`, which a Worker secret could not provide (secrets are read-only at
+  runtime).
+- **Hashing:** SHA-256 of the code (codes are high-entropy, so a fast hash is
+  fine); compare in constant time.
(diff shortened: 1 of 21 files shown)


docs/superpowers/specs/2026-06-21-passkey-indieauth-design.md
```diff
@@ -0,0 +1,255 @@
...
+(see Open decisions §1).
+
+> Caveat to record: `@dwk/webauthn` is self-described "exploratory, lowest
+> priority" — the least-mature package in the set. We are putting the owner's
+> primary identity on it. Acceptable given the owner controls the whole stack and
+> can reset auth (§ Recovery), but worth stating plainly.
+
+## Architecture
+
+Everything mounts under `/auth` in `site-entry.js`, ahead of the `ASSETS`
```
davidwkeith, line 67: **Security: consent token in a URL query parameter leaks in logs and history.**

The redirect `302 → /auth?...&_consent=<token>` puts the HMAC token in:
- Cloudflare access logs (default, cannot be disabled per-URL)
- Browser history
- `Referer` header sent to the IndieAuth client's `redirect_uri` on the next navigation

Even though the token is short-lived and HMAC-bound, normalising URL-embedded tokens is a pattern we're better off avoiding here.

Suggested alternative: after the `/auth/consent` POST succeeds, set a short-lived `__anglesite_consent` HttpOnly cookie containing the HMAC, then `302 → /auth?...original params only...`. The `approveAuthorization` hook reads the cookie instead of `_consent` from the URL, verifies it, and clears it (by setting `Max-Age=0` on the approval `Response` headers... actually `AuthorizationApproval` doesn't return a `Response`, so the library handles that). If clearing the cookie from an `AuthorizationApproval` return path isn't possible, use a one-shot nonce stored in `OWNER_AUTH_DB` instead — write on mint, delete on consume, so replay is impossible regardless of expiry.

---

docs/superpowers/specs/2026-06-21-passkey-indieauth-design.md
```diff
@@ -0,0 +1,255 @@
...
+fallthrough, gated on `env.AUTH_DB` (as today):
+
+```
+/auth                       → @dwk/indieauth authorization + token + metadata
+/auth/webauthn/register/*    → @dwk/webauthn (owner-only gated, see §1)
+/auth/webauthn/authenticate/* → @dwk/webauthn (open — proves possession only)
+/auth/register               → owner-only registration page (HTML)
```
davidwkeith, line 74: **`nonce` in the HMAC input is undefined — needs specification.**

The token is described as `HMAC over client_id|redirect_uri|scope|nonce|exp` but the spec never says where `nonce` comes from. Options:
- The PKCE `state` parameter from the original request (already bound to the client's session, good choice)
- A separately generated random value stored server-side (more entropy but requires storage)
- The PKCE `code_challenge` (already present in the request, cryptographically tied to the verifier)

Using `state` is probably the right call (it's already per-request and client-generated), but the implementation plan needs to be explicit. If `state` is absent or optional in a particular IndieAuth client, what's the fallback?

---

docs/superpowers/specs/2026-06-21-passkey-indieauth-design.md
```diff
@@ -0,0 +1,255 @@
...
+`rpId`, `origin`) that derives from `env.SITE_URL`.
+
+### The flow
+
+```
+1. Client → GET /auth?response_type=code&client_id&redirect_uri&state
+            &code_challenge&code_challenge_method=S256&scope&me
+2. @dwk/indieauth validates → approveAuthorization(request, httpRequest)
+3. our hook reads the signed __anglesite_owner session cookie:
+   a. no/expired session →
```
davidwkeith, line 89: **Is the 15-minute session a hard cap or does it slide?**

15 minutes is aggressive if the owner grants consent to multiple IndieAuth clients in a sitting (e.g., setting up a new Micropub client, then an RSS reader, then a commenting service). Specify whether:
- The cookie is refreshed on each successful `approveAuthorization` call (sliding window), or
- 15 min is a hard cap requiring a new passkey assertion each time.

If it's a hard cap, document that explicitly — it's a usability trade-off the implementor should understand. If it's sliding, the refresh logic needs to be in the spec.

---

docs/superpowers/specs/2026-06-21-passkey-indieauth-design.md
```diff
@@ -0,0 +1,255 @@
...
+
+### Session
+
+A signed cookie `__anglesite_owner`, reusing the existing membership HMAC helper
+shape in `site-entry.js` (`verifyMembershipCookie`/`crypto.subtle` HMAC-SHA-256).
+Payload `{ sub: "owner", iat, exp }`, `exp` ~15 min, `HttpOnly; Secure; SameSite=Lax`.
+A fresh passkey assertion re-issues it. Signing key: a new
+`INDIEAUTH_SESSION_KEY` secret (or reuse `INDIEAUTH_SIGNING_KEY`).
+
+### Registration
```
davidwkeith, line 120: **Backup code hashing: "high-entropy" needs to be quantified, and bare SHA-256 may be insufficient.**

"Codes are high-entropy, so a fast hash is fine" is only true if the codes are long enough that the hash output space can't be exhausted. The spec doesn't state the code format (length, alphabet, entropy bits). The implementation will need to define this — please add it here.

Recommended: 10 codes × 8 random bytes each = 64 bits entropy per code, base32-encoded → `XXXX-XXXX-XXXX` format. At that entropy level bare SHA-256 is defensible on a DB leak.

However, a cleaner option at essentially zero cost: **HMAC-SHA-256 keyed with a Worker secret** (e.g., `OWNER_AUTH_DB_KEY`). Keyed hashing means a leaked D1 table alone reveals nothing — the attacker also needs the Worker secret. Worth the extra secret.

Also: the rate-limit bullet says "per session/IP" — session is stateful (cookie), IP is a header Cloudflare sets. Specify which D1 table tracks attempts and the limit (e.g., 5/hour per IP, 10 total per session).

---

docs/superpowers/specs/2026-06-21-passkey-indieauth-design.md
```diff
@@ -0,0 +1,255 @@
...
+
+### Backup codes
+
+A second recovery factor alongside multiple passkeys and the redeploy root. At
+setup (and regenerable from an authenticated session), the worker generates **10
+single-use backup codes**, shows them once for the owner to print/save, and
+stores only their hashes. The consent page offers "Use a backup code instead of a
+passkey"; a submitted code is hashed and matched against an unused stored hash —
+on match it establishes the owner session and the code is marked consumed. A used
+or unknown code returns the consent page unchanged (no oracle).
```
davidwkeith, line 136: **`INDIEAUTH_SESSION_KEY` row contradicts the Resolved decision §3.**

The bindings table says `"new (owner session cookie) — or reuse above"` but §3 Resolved says `"a separate new INDIEAUTH_SESSION_KEY"`. The "or reuse above" hedge should be removed. It's already been decided, and leaving the ambiguity here means an implementor might reasonably pick the wrong path. Drop the "or reuse above" clause.

---

docs/superpowers/specs/2026-06-21-passkey-indieauth-design.md
```diff
@@ -0,0 +1,255 @@
...
+  runtime).
+- **Hashing:** SHA-256 of the code (codes are high-entropy, so a fast hash is
+  fine); compare in constant time.
+- **Rate limit:** cap backup-code attempts per session/IP to blunt guessing,
+  even against high-entropy codes.
+
+## Bindings & wrangler
+
+| Binding | Type | Source |
+|---|---|---|
```
davidwkeith, line 151: **`INDIEWEB_REG_TOKEN` has no stated expiry — add one.**

The skill generates the token and shows the owner a link once. Between that moment and when the owner visits the link, the token is live indefinitely (until the first credential enrolls and the skill rotates/clears it). If the owner never clicks the link, the window stays open forever.

Add an expiry: e.g., the skill sets the secret to `<random>:<issued_unix_ts>` and the handler rejects it after 1 hour (or 24 h for timezone flexibility). This is a low-cost addition that closes the bootstrap window without requiring any new storage.

---

docs/superpowers/specs/2026-06-21-passkey-indieauth-design.md
```diff
@@ -0,0 +1,255 @@
...
+| `AUTH_DB` | D1 | already declared (indieauth codes/tokens) |
+| `WEBAUTHN` | **Durable Object namespace** → `WebAuthnObject` | **new** |
+| `SITE_URL` | var | already added (#370); also drives `rpId`/`origin` |
+| `INDIEAUTH_SIGNING_KEY` | secret | already (token signing) |
+| `INDIEAUTH_SESSION_KEY` | secret | **new** (owner session cookie) — or reuse above |
+| `INDIEWEB_REG_TOKEN` | secret | **new** (registration bootstrap, §1) |
+| `OWNER_AUTH_DB` | D1 | **new** (single-use backup-code hashes) |
+
```
davidwkeith, line 159: **`residentKey` policy is unspecified — required for a real passkey experience.**

The spec mandates `user_verification: "required"` but doesn't set `residentKey`. Without `residentKey: "required"` (or `"preferred"`) in the registration options, some authenticators create non-discoverable (server-side) credentials. That means `/authenticate/options` needs a username hint — which defeats the "owner just taps the passkey" UX and breaks the flow on the consent page.

For `@dwk/webauthn`, check whether `residentKey` is configurable or baked in. If it's configurable, the spec should require `"required"`. If it's baked, note that here.

---

docs/superpowers/specs/2026-06-21-passkey-indieauth-design.md
```diff
@@ -0,0 +1,255 @@
...
+The `WEBAUTHN` DO is the **first Durable Object in the Anglesite template**. It
+needs a `durable_objects.bindings` entry **and** a `migrations` block in
+`wrangler.jsonc`:
+
+```jsonc
```
davidwkeith, line 164: **`workers.dev` preview origin is a footgun — add a call-out.**

The spec says passkeys are domain-bound and the skill requires a custom domain. But during development, `wrangler dev` and `wrangler deploy --env preview` serve from `*.workers.dev`. A developer who registers credentials against `mysite.workers.dev` will find them silently rejected on `mysite.com` (origin mismatch), and vice versa.

Worth adding a sentence: "Passkey registration and authentication are bound to the `rpId`/`origin` derived from `SITE_URL`. Credentials enrolled in a preview/workers.dev environment are not usable on the production custom domain — always use the production `wrangler.jsonc` when bootstrapping the owner passkey."

---

docs/superpowers/specs/2026-06-21-passkey-indieauth-design.md
```diff
@@ -0,0 +1,255 @@
...
+
+- `approveAuthorization`: no session → returns a `Response` (the consent page),
+  not an approval; never mints on an unauthenticated request.
+- valid session + matching consent token → returns `AuthorizationApproval` with
+  `me === SITE_URL` and the requested scopes.
+- forged/missing/mismatched `_consent` token with a valid session → still returns
+  the consent page, never an approval (deep-link-skip blocked).
+- `/auth/webauthn/register/*` with neither token nor session → 403.
+- backup code: an unused code establishes a session and is marked `used_at`; the
+  same code a second time fails; an unknown code returns the consent page
```
davidwkeith, line 229: **Build step 1 leaves registration endpoints ungated — needs an explicit deployment warning.**

Step 1 wires the Durable Object and makes the `/register/*` ceremonies "reachable (no gating yet)." If someone deploys the site at this intermediate step — CI auto-deploy, accidental `wrangler deploy` — any visitor could enrol a credential and become the `@dwk/webauthn` RP's only registered user.

Please add: "**Do not deploy to production between steps 1 and 3.** Step 1 only proves the DO boots in a local or preview environment; registration gating is added in step 3." Alternatively, invert the sequence: add the registration gate in step 1 as a hard 403 (checking for a token or session that doesn't exist yet is trivially "all fail"), and relax it in step 3 when the bootstrap flow is wired.

---

davidwkeith (commented):
Good spec overall — the contract with `approveAuthorization`, the three-layer recovery model, and the TDD plan are all solid. Eight issues to address before this approves, ranging from one real security concern to a few gaps that would bite the implementor.

**Must-fix before implementation starts:**

1. **Consent token in URL** (lines 59-67) — The `_consent` token landing in a GET query string leaks into Cloudflare logs, browser history, and `Referer` headers. Move it to a short-lived HttpOnly cookie set on the `/auth/consent` POST response, cleared on first use. See inline comment for two concrete alternatives.

2. **`nonce` in the consent HMAC is undefined** (lines 73-74) — The HMAC covers `client_id|redirect_uri|scope|nonce|exp` but `nonce` is never defined. Is it the PKCE `state`? A separately generated value? Needs to be specified.

3. **Build step 1 deploys ungated registration** (lines 228-229) — "ceremonies reachable (no gating yet)" is a production footgun if CI picks it up. Either add an explicit "do not deploy to production between steps 1 and 3" callout, or flip the order and gate-first.

**Should fix before merging:**

4. **Backup code format unspecified; bare SHA-256 may be insufficient** (lines 118-120) — Specify the code length/alphabet/entropy so "high-entropy" is verifiable. HMAC-SHA-256 keyed with a Worker secret is a low-cost improvement over bare SHA-256 if the D1 table ever leaks.

5. **`INDIEWEB_REG_TOKEN` has no expiry** (lines 148-151) — Add a timestamp to the secret and reject it after 1 hour. Low-cost, closes the bootstrap window.

6. **`INDIEAUTH_SESSION_KEY` row says "or reuse above"** (lines 135-136) — Contradicts §3 Resolved. Drop the hedge.

7. **`residentKey` policy missing** (lines 157-159) — Without `residentKey: "required"`, some authenticators create non-discoverable credentials and break the usernameless assertion flow on the consent page.

8. **`workers.dev` origin footgun** (lines 163-164) — Passkey credentials enrolled on `*.workers.dev` won't work on the production custom domain. Add a callout.

**Minor:**

- Session duration (line 88): state whether the 15-min cookie slides on use or is a hard cap. Either is fine, but the implementor needs to know.


---

PR #372, opened 2026-06-21
fix(indieweb): wire Webmention to the real @dwk API + fix stored XSS (#363)

davidwkeith:
Re-platforms the Webmention path onto the published `@dwk/webmention` API, fixes the boot crash the 1.1.0 Worker would hit against the real packages, and closes the stored-XSS finding. Addresses **findings 2 & 3** of #363 (finding 1 is in the `@dwk/*` package repos; IndieAuth + Micropub are deferred to a follow-up).

## Why
`worker/site-entry.js` is every site's Worker `main`. In 1.1.0 it statically imported `@dwk/*` and constructed the handlers at module load with bare config — which **throws on boot** against the published packages (`createWebmention()`/`createIndieAuth()` read `config.baseUrl`/`config.logger` on `undefined`). The scaffold only appeared to build because `astro build` doesn't bundle the Worker and the tests mocked the packages.

## What changed
- **Security (finding 3):** `renderMention()` interpolated external webmention `author_url` / `author_photo` / `url` into `href`/`src` with only `escapeHtml()`, which doesn't block `javascript:`/`data:` schemes. URLs now pass a `safeUrl()` http(s) allowlist; anchors get `rel="nofollow ugc noopener"`.
- **Re-wire (finding 2):** removed module-load construction (fixes the boot crash). IndieAuth + Micropub routes report **501** pending their auth follow-up. Webmention moves to a new injectable **`worker/webmention.js`** composed against `createWebmention` + `createWebmentionQueueConsumer`, with a **custom `InboxStore`** that re-fetches each verified source (SSRF-guarded via `safeFetch`) and parses **microformats2** to persist the author h-card, content, permalink, and reply/like/repost type — the package's default inbox only stores `(source, target, verified_at)`.
- **Architecture:** the module's `@dwk/webmention` + `microformats-parser` imports can't ship for all sites, so `site-entry.js` ships four `/* @anglesite-inject:* */` sentinels that `/anglesite:indieweb` fills in (new **Step 4b**: install deps + replace sentinels) only when Webmention is enabled. Plain sites never install or bundle the deps.
- **Skill:** truthful availability (Webmention live; IndieAuth/Micropub 501-pending), install/injection steps, auto-created schema docs, and the `TOKEN_SIGNING_KEY` correction for the IndieAuth follow-up.

## Tests
- New `tests/webmention-module.test.ts` (14) — real `mf2` parse (reply/like/repost/mention), receiver + queue-drain gating, edge-render gating.
- `tests/webmention-render-xss.test.ts` (8) — `safeUrl`/`renderMention` scheme hardening.
- `tests/indieweb-dispatch.test.ts` — rewritten for the 501 default, webmention-unwired passthrough, and the injection-sentinel contract.
- Full suite: **2658 pass**; only the pre-existing `optimize-images-packaging` e2e fails (confirmed failing on a clean tree, sharp/env-related). Injection recipe validated to produce valid JS.

## Follow-ups (not in this PR)
1. Wire IndieAuth (`approveAuthorization` consent endpoint + `TOKEN_SIGNING_KEY`) and Micropub — removes the 501 stubs; the `dwk-indieauth`/`dwk-micropub` test stubs (still exporting the old `createHandler`, currently unused) get rewritten then.
2. Fix the extensionless ESM re-exports (finding 1) and republish in the `@dwk/*` repos.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "anglesite",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "anglesite",
-      "version": "1.1.0",
+      "version": "1.2.0",
       "license": "ISC",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.28.0",
@@ -16,6 +16,7 @@
         "@types/semver": "^7.7.1",
         "html-validate": "^10.11.2",
         "markdownlint-cli2": "^0.22.0",
+        "microformats-parser": "^2.0.4",
         "robots-parser": "^3.0.1",
         "schema-dts": "^2.0.0",
         "semver": "^7.7.4",
@@ -3672,6 +3673,19 @@
         "node": ">= 8"
       }
     },
+    "node_modules/microformats-parser": {
+      "version": "2.0.4",
+      "resolved": "https://registry.npmjs.org/microformats-parser/-/microformats-parser-2.0.4.tgz",
+      "integrity": "sha512-DA2yt3uz2JjupBGoNvaG9ngBP5vSTI1ky2yhxBai/RnQrlzo+gEzuCdvwIIjj2nh3uVPDybTP5u7uua7pOa6LA==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "parse5": "^7.1.2"
+      },
+      "engines": {
+        "node": ">=18"
+      }
+    },
     "node_modules/micromark": {
       "version": "4.0.2",
       "resolved": "https://registry.npmjs.org/micromark/-/micromark-4.0.2.tgz",
@@ -4432,6 +4446,32 @@
       },
       "funding": {
         "url": "https://github.com/sponsors/sindresorhus"
+      }
+    },
+    "node_modules/parse5": {
+      "version": "7.3.0",
+      "resolved": "https://registry.npmjs.org/parse5/-/parse5-7.3.0.tgz",
+      "integrity": "sha512-IInvU7fabl34qmi9gY8XOVxhYyMyuH2xUNpb2q8/Y+7552KlejkRvqvD19nMoUW/uQGGbqNpA6Tufu5FL5BZgw==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "entities": "^6.0.0"
+      },
+      "funding": {
+        "url": "https://github.com/inikulin/parse5?sponsor=1"
+      }
+    },
+    "node_modules/parse5/node_modules/entities": {
+      "version": "6.0.1",
+      "resolved": "https://registry.npmjs.org/entities/-/entities-6.0.1.tgz",
+      "integrity": "sha512-aN97NXWF6AWBTahfVOIrB/NShkzi5H7F9r1s9mD3cDj4Ko5f2qhhVoYMibXF7GlLveb/D2ioWay8lxI97Ven3g==",
+      "dev": true,
+      "license": "BSD-2-Clause",
+      "engines": {
+        "node": ">=0.12"
+      },
+      "funding": {
+        "url": "https://github.com/fb55/entities?sponsor=1"
       }
     },
     "node_modules/parseurl": {
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "@types/semver": "^7.7.1",
     "html-validate": "^10.11.2",
     "markdownlint-cli2": "^0.22.0",
+    "microformats-parser": "^2.0.4",
     "robots-parser": "^3.0.1",
     "schema-dts": "^2.0.0",
     "semver": "^7.7.4",
--- a/skills/indieweb/SKILL.md
+++ b/skills/indieweb/SKILL.md
@@ -55,19 +55,25 @@ Do not attempt git installs, forks, or workarounds. The gate is intentional.
 
 ## Step 2 — Choose endpoints
 
-Tell the owner: "There are three IndieWeb endpoints you can enable — each works independently, and I'd recommend all three for the full experience:"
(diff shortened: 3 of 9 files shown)


davidwkeith (commented):
## Code Review — PR #372: Webmention replatform + stored-XSS fix

Good foundational work here. The sentinel injection architecture is clean, the XSS fix is correct, and the test coverage is thorough. There are four issues worth addressing before merge, ordered by severity.

---

### 🔴 Bug: `ensureSchema` caches rejected Promises forever

**`template/worker/webmention.js`, `createWebmentionInbox`:**

```js
let ready = null;
const ensureSchema = () => {
  ready ??= db.prepare(CREATE_TABLE_SQL).run().then(() => undefined);
  return ready;
};
```

`??=` only assigns when `ready` is `null` or `undefined`. If the first `CREATE TABLE` call hits a transient D1 error, `ready` becomes a rejected Promise and stays one — every subsequent call to `store()`, `remove()`, or `list()` will reject without ever retrying. A single D1 hiccup during the first webmention permanently disables the inbox for that isolate's lifetime.

Fix: clear `ready` on rejection so the next call retries:

```js
const ensureSchema = () => {
  if (!ready) {
    ready = db.prepare(CREATE_TABLE_SQL).run()
      .then(() => undefined)
      .catch(err => { ready = null; throw err; });
  }
  return ready;
};
```

---

### 🔴 Security: `rel` is missing `noreferrer`

**`template/worker/webmention.js`, `renderMention()`:**

```js
`<a … rel="nofollow ugc noopener" href="…">`
```

`noopener` prevents the new tab from accessing `window.opener`, but the browser still sends the `Referer` header when a visitor navigates to the external URL. For webmention authors — arbitrary third-party sites — that leaks the URL of the note the mention is displayed on to the linked domain.

`noreferrer` implies `noopener` and suppresses the header. The correct value:

```js
rel="nofollow ugc noopener noreferrer"
```

Both anchors (author URL and permalink) need this change.

---

### 🟡 Correctness: NULL ordering in `ORDER BY published ASC`

**`template/worker/webmention.js`, `loadWebmentions()`:**

```sql
ORDER BY published ASC
```

`published` is a nullable `TEXT` column. SQLite sorts NULLs before all other values ascending, so any mention whose source page had no `dt-published` will float to the top of every thread above dated replies — opposite of chronological intent.

Prefer:

```sql
ORDER BY COALESCE(published, verified_at) ASC
```

`verified_at` is `INTEGER NOT NULL`, so this degrades gracefully for undated mentions while preserving chronological order.

---

### 🟡 Skill robustness: no idempotency guard on sentinel injection

**`skills/indieweb/SKILL.md`, Step 4b:**

The injection step replaces four `/* @anglesite-inject:* */` comments with live code. If the owner runs `/anglesite:indieweb` a second time (e.g., to eventually add IndieAuth when it lands), the skill will look for sentinels that no longer exist. The Edit tool will error on "string not found," leaving the worker in a partially-broken state.

Step 4b should start with:

> Before replacing each sentinel, confirm it still exists in `worker/site-entry.js`. If a sentinel is already replaced (the `import` or dispatch line is present), skip it — Webmention is already wired.

This also naturally skips the `npm install` step on re-runs.

---

### ℹ️ Minor observations (no changes needed)

- **`receivers` Map** — unbounded in theory, bounded to ~2 entries per isolate in practice (main domain + preview). Fine.
- **Bare h-card handling in `parseMention`** — adapting a standalone h-card into `{ author: [card] }` for aggregator sources is a nice spec-faithful touch.
- **`webmention.scheduled` removal** — the rationale comment ("no scheduled arm — verification is queue-driven") is clear. Confirmed correct by the stub removing the `scheduled` mock.
- **Sentinel contract test** — asserting all four `@anglesite-inject:*` anchors are present in the shipped source is a good skill-contract guard.

The XSS fix itself (`safeUrl` + `escapeHtml` layering, `new URL()` protocol check catching even `JavaScript:` via normalization) is correct and the regression tests are comprehensive.

---

davidwkeith:
Thanks for the review — all four addressed in bd8da20.

| # | Item | Resolution |
|---|---|---|
| 🔴 | `ensureSchema` caches rejected promises | `ready` is now cleared in a `.catch` before rethrowing, so a transient first-write error retries instead of wedging the isolate. Added a regression test (`createWebmentionInbox() — schema resilience`) that rejects the first `CREATE TABLE` and asserts the second `store()` retries and succeeds. |
| 🔴 | `rel` missing `noreferrer` | Both anchors are now `rel="nofollow ugc noopener noreferrer"`. Updated the XSS test assertion. |
| 🟡 | NULL ordering in `ORDER BY published` | Fixed — with one refinement. A raw `COALESCE(published, verified_at)` still misorders because `published` is TEXT and `verified_at` is INTEGER, so SQLite's type affinity sorts *all* undated rows before *all* dated ones. Normalizing the fallback to an ISO string keeps the key uniformly text and chronological: `ORDER BY COALESCE(published, datetime(verified_at / 1000, 'unixepoch')) ASC`. |
| 🟡 | No idempotency guard on injection | Step 4b now begins by checking for `import { handleWebmention` in `site-entry.js` and skips both the install and the sentinel replacements if Webmention is already wired — so a re-run (e.g. to add IndieAuth later) won't error on missing sentinels. |

Full suite: 2659 pass (only the pre-existing `optimize-images-packaging` e2e fails, unrelated/sharp-env).

davidwkeith:
Superseded by work merged to `main` while this was open:
- **#369** — webmention render XSS fix (this PR's finding 3)
- **#370** — realign Webmention to the real `@dwk/webmention` API (finding 2)
- **#371** — passkey-backed IndieAuth (the follow-up this PR deferred)

Main's implementation uses a different architecture (static `@dwk/*` imports + `SITE_URL` + `WEBMENTION_INBOX` + the package's default `createD1Inbox`), so this branch can't be merged without regressing #370/#371. The one piece not on main is the **rich author-card render** (mf2 parsing) — I'm reimplementing just that delta on top of main's architecture in a fresh branch. Closing this in favor of that.

PR #373, opened 2026-06-21
feat(indieweb): rich Webmention author cards via mf2-parsing inbox

davidwkeith:
Builds on `main`'s Webmention realignment (#370) to render **full mention cards** — sender name, photo, content snippet, and reply/like/repost type — instead of just the source host. This is the rich-cards delta from the now-closed #372, reimplemented on main's architecture.

## Why
`@dwk/webmention`'s default `createD1Inbox` persists only `(source, target, verified_at)`, so #370's edge-render can show no more than the mentioning host. Rich cards need the source's author h-card and content, which means parsing the source's microformats2 at verification time.

## What changed
- **New `worker/webmention-inbox.js`** — `createRichWebmentionInbox(db)` is a drop-in `InboxStore` (`store`/`remove`/`list`, the beta.2 contract). On `store` it re-fetches the verified source (SSRF-guarded via the package's `safeFetch`) and parses its mf2 (`microformats-parser`) into an extended `webmentions` table: author name/url/photo, content, permalink, type.
- **`site-entry.js`** — `webmentionFor(env)` builds the rich inbox and passes it to **both** the queue consumer (so verification writes the rich row) and the edge-render reader. `renderMention` becomes a full card (photo, author link, content, permalink, `webmention-<type>` class). Every external URL is still `safeUrl`-checked; a mention with no usable http(s) URL is dropped — the XSS contract from #369 is preserved (main's `indieweb-render-safety.test.ts` passes unchanged).
- **Resilience** — `ensureSchema` clears its cached promise on failure (transient first-write D1 error retries instead of wedging the inbox); chronological ordering is NULL-safe via `COALESCE(published, datetime(verified_at/1000,'unixepoch'))`.
- **Skill** — installs `microformats-parser` alongside `@dwk/webmention` (filling a missing install step), documents the rich schema, and gives the drop-table migration from the minimal install.

## Tests
- `tests/webmention-inbox.test.ts` — real mf2 parse (reply/like/repost/fallback), `store`→fetch→parse→upsert, fetch-failure degrade, schema-retry, rich `list`.
- `tests/webmention-render-rich.test.ts` — author-card output + per-field URL safety (`javascript:`/`data:` dropped) + type clamp + no-URL drop.
- `tests/indieweb-dispatch.test.ts` — edge-render mock moved from the stub inbox to a `prepare()`-based D1 fake (the rich inbox reads via `db.prepare().all()`).
- Full suite: **2693 pass**; only the pre-existing `optimize-images-packaging` e2e fails (sharp/env, confirmed failing on a clean tree).

## Note
`site-entry.js` (on main, unmodified) imports `readCookie` from `owner-auth.js` *and* declares it locally — esbuild tolerates it by shadowing, but it's a latent duplicate binding (the import is dead). Flagging since I noticed it; not touched here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "@types/semver": "^7.7.1",
     "html-validate": "^10.11.2",
     "markdownlint-cli2": "^0.22.0",
+    "microformats-parser": "^2.0.4",
     "robots-parser": "^3.0.1",
     "schema-dts": "^2.0.0",
     "semver": "^7.7.4",
--- a/template/worker/site-entry.js
+++ b/template/worker/site-entry.js
@@ -59,8 +59,8 @@ import { createHandler as createMicropub } from "@dwk/micropub";
 import {
   createWebmention,
   createWebmentionQueueConsumer,
-  createD1Inbox,
 } from "@dwk/webmention";
+import { createRichWebmentionInbox } from "./webmention-inbox.js";
 import { createWebAuthn, WebAuthnObject } from "@dwk/webauthn";
 import { sync as syncMicropubBridge } from "./indieweb-bridge.js";
 import {
@@ -192,10 +192,17 @@ function webmentionFor(env) {
           "to configure it.",
       );
     }
+    // The rich inbox parses each verified source's microformats2 so mentions
+    // carry the author h-card + content, not just the source/target pair. It
+    // backs BOTH the queue consumer (verification writes the rich row) and the
+    // edge-render reader (the page shows it).
+    const inbox = env.WEBMENTION_INBOX
+      ? createRichWebmentionInbox(env.WEBMENTION_INBOX)
+      : null;
     bundle = {
       handler: createWebmention({ baseUrl }),
-      consumer: createWebmentionQueueConsumer({ baseUrl }),
-      inbox: env.WEBMENTION_INBOX ? createD1Inbox(env.WEBMENTION_INBOX) : null,
+      consumer: createWebmentionQueueConsumer({ baseUrl, inbox }),
+      inbox,
     };
     webmentionByEnv.set(env, bundle);
   }
@@ -430,7 +437,7 @@ const WEBMENTION_TARGETS_TTL_MS = 60_000;
 const webmentionTargetsCache = new WeakMap();
 
 // `dbKey` is the WEBMENTION_INBOX binding (the stable cache key); `inbox` is the
-// createD1Inbox reader over it. Each isolate runs the full list() scan at most
+// rich inbox reader over it. Each isolate runs the full list() scan at most
 // once per TTL — every other note/post request is an in-memory set lookup. A
 // failure serves the stale (or empty) set rather than breaking the render.
 async function loadWebmentionTargets(dbKey, inbox) {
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -29,6 +29,7 @@
     "@types/semver": "^7.7.1",
     "html-validate": "^10.11.2",
     "markdownlint-cli2": "^0.22.0",
+    "microformats-parser": "^2.0.4",
```
davidwkeith, line 32: `microformats-parser` is added here as a plugin devDependency (for running the `parseMention` tests in this repo), but `template/worker/webmention-inbox.js` imports it at the top level, meaning every scaffolded site needs it resolvable at `wrangler build` time.

`template/package.json` isn't updated in this PR. If `@dwk/webmention` and `microformats-parser` are intentionally installed only by the `/anglesite:indieweb` skill step (not pre-scaffolded), then a site that was set up with webmention before this change — and deploys again without re-running the skill — will get a build failure from the unresolved `microformats-parser` import.

Either:
1. Add `microformats-parser` to `template/package.json` so all scaffolded sites have it after `npm install`, or
2. Make the import inside `webmention-inbox.js` dynamic (inside `store()`) so esbuild only needs it when the file is actually used, or
3. Confirm that `template/package.json` already lists it (in which case the diff is incomplete).

This is the one blocker I see — everything else is sound.

---

template/worker/webmention-inbox.js
```diff
@@ -0,0 +1,250 @@
...
+ * parses its microformats2, and persists the author h-card, content, permalink,
+ * and reply/like/repost type to an extended `webmentions` table that
+ * `renderMention()` reads to build full mention cards.
+ *
+ * It is wired in `site-entry.js`'s `webmentionFor(env)` as BOTH the queue
+ * consumer's inbox (so verification populates the rich row) and the edge-render
+ * reader (so the page shows it). A fetch/parse failure degrades to a
+ * source/target-only row rather than dropping the verified mention.
+ */
+import { safeFetch } from "@dwk/webmention";
```
davidwkeith: The project style guide (CLAUDE.md) says "Don't write multi-paragraph docstrings or multi-line comment blocks — one short line max." This 17-line JSDoc explains what the module does, which the module name and code already make clear. Worth trimming to the one non-obvious thing: that the same inbox instance must be passed to both the consumer and the reader so verification and edge-render share the same extended schema.

---

template/worker/webmention-inbox.js
```diff
@@ -0,0 +1,250 @@
...
+      let parsed = { type: "mention", url: mention.source };
+      try {
+        const { response } = await safeFetch(
+          fetchImpl,
+          mention.source,
+          { headers: { accept: "text/html" } },
+          { timeoutMs: 8000 },
+        );
+        if (response.ok) {
+          parsed = { ...parsed, ...parseMention(await response.text(), mention) };
```
davidwkeith, line 88: 8 seconds is generous for a queue consumer, but worth confirming it fits inside the package's own queue-consumer timeout budget. If `safeFetch` internally enforces a shorter limit anyway, the `8000` here is redundant. If not, slow webmention sources could stall the queue drain — consider 3–5 s as a safer default.

---

template/worker/site-entry.js
```diff
@@ -481,17 +488,36 @@ class WebmentionInjector {
...
+  const authorPhoto = safeUrl(m.author_photo);
+  // safeUrl returns normalized, valid URL strings, so re-parsing for a display
+  // host cannot throw.
+  const name = escapeHtml(
+    m.author_name ||
+      (authorUrl ? new URL(authorUrl).host : new URL(permalink).host),
   );
+  const type = MENTION_TYPES.has(m.type) ? m.type : "mention";
+
+  const photo = authorPhoto
```
davidwkeith, line 510: The photo URL passes `safeUrl` (scheme-checked, no XSS risk), but loading it without a `referrerpolicy` leaks the visitor's current page URL as a `Referer` header to the avatar's origin server — which is attacker-controlled. Add `referrerpolicy="no-referrer"` to the `` tag:

```js
``'&lt;img class="u-photo" src="${escapeHtml(authorPhoto)}" alt="" width="48" height="48" loading="lazy" referrerpolicy="no-referrer" /&gt;'``
```

---

tests/webmention-inbox.test.ts
davidwkeith, line 1: The `remove(source, target)` method has no test. It's a one-liner `DELETE WHERE source = ?1 AND target = ?2`, but a quick test would confirm the prepared-statement binding order matches the signature — easy to transpose.

---

davidwkeith (commented):
## Overall

Clean, well-reasoned PR. The architecture is right — same inbox instance backing both the queue consumer and the edge-render reader, `safeUrl` applied to every external field before it reaches an `href`/`src`, type allowlist to prevent CSS class injection, graceful degradation on fetch failure, schema-retry to avoid a wedged isolate. The tests are substantive.

One blocker and three smaller items to address before merge.

### Blocker

**`template/package.json` not updated.** `webmention-inbox.js` is now scaffolded as a template file and `site-entry.js` imports it statically, so `microformats-parser` must be resolvable at `wrangler build` time for every site — not just those that have run `/anglesite:indieweb`. Details on the options in the inline comment on `package.json`.

### Smaller items

- **`referrerpolicy` on the author photo ``** — avatar origin is attacker-controlled; omitting it leaks the visitor's page URL. Inline comment on `site-entry.js`.
- **8 s fetch timeout** — might be longer than needed; could stall queue drain on slow sources. Inline comment on `webmention-inbox.js`.
- **Multi-paragraph JSDoc header** — conflicts with the style guide. Inline comment on `webmention-inbox.js`.
- **`remove()` not tested** — minor, but easy to add. File-level comment on the test.

---

davidwkeith:
All five addressed in 849663e.

| Item | Resolution |
|---|---|
| 🔴 **`template/package.json` not updated** | Added `microformats-parser@2.0.4` to `template/package.json` and regenerated `template/package-lock.json` (verified `npm ci` in sync, so the `template-lockfile` job stays green). Since `webmention-inbox.js` is scaffolded and `site-entry.js` imports it statically, it's a hard build dep for **every** site — option 1. (Option 2/dynamic-import doesn't actually help: esbuild bundles dynamic imports too, so the package still has to resolve at build.) |
| `referrerpolicy` on `` | Added `referrerpolicy="no-referrer"` to the author-photo img; asserted in the rich-render test. |
| 8s timeout | Lowered to **5s**, under the package's 10s `DEFAULT_TIMEOUT_MS`, so one slow source can't stall the in-sequence queue drain. Added a comment noting why. |
| Multi-paragraph JSDoc | Trimmed the 17-line header to a single line (kept the one non-obvious bit: wire the *same* instance into both the consumer and the reader). |
| `remove()` untested | Added a test asserting the `DELETE` binds `(source, target)` in order. |

Full suite: 2694 pass (only the pre-existing `optimize-images-packaging` e2e fails — sharp/env).

**One thing to flag on the blocker:** `microformats-parser` is now a baseline `template/` dep, so it lands in **every** scaffolded site's `node_modules` and Worker bundle (parse5 included), even ones that never enable Webmention — because `site-entry.js` imports `webmention-inbox.js` unconditionally. That's consistent with how `@dwk/webmention` is already statically imported there, but note `@dwk/*` are *not* in `template/package.json`, so a plain site's `wrangler build` would still fail on those unresolved imports. If the intent is for plain sites to build clean, that's a broader call (gate the webmention imports behind the skill, or add `@dwk/*` to `template/`) — happy to follow up whichever way you want.

Issue #365, opened 2026-06-18
pre-deploy-check.sh: two false positives block legit deploys (URL/DOI matched as phone; same-origin script flagged as third-party)


davidwkeith:
## Summary

`scripts/pre-deploy-check.sh` (the `PreToolUse` git-push gate) produces **two false positives** that block legitimate deploys to `main`. Both checks are stricter/cruder than their counterparts in `scripts/pre-deploy-check.ts`, which handle the same cases correctly. The `.sh` gate and the `.ts` scanner have drifted.

Observed on `anglesite@1.1.0` (also present in `1.0.0-beta.7`), real site: `pulletsforever.com`.

```
Deploy blocked: Possible phone number found in built HTML; Unauthorized third-party script tag found;
```

Neither finding was real.

---

## Bug 1 — Phone PII check matches digit runs inside URLs / DOIs

`pre-deploy-check.sh`:

```sh
PHONE_HITS=$(grep -rE '\(?\d{3}\)?[-.[[:space:]]]?\d{3}[-.[[:space:]]]?\d{4}' "$DIST/" --include='*.html' ...)
```

This unanchored pattern matches digit groups embedded in longer strings. A Nature article link in a blog post:

```
https://www.nature.com/articles/s41598-025-97652-6
```

matches `598-025-9765` → reported as "Possible phone number found in built HTML".

**Root cause:** no boundary guards. The TS scanner already solved this — `scripts/pre-deploy-check.ts`:

```ts
export const phonePattern = /(?<![\d.-])\(?\d{3}\)?[-.\s]?\d{3}[-.\s]?\d{4}(?![\d-])/g;
```

The negative lookbehind `(?<![\d.-])` and lookahead `(?![\d-])` reject matches embedded in longer digit/hyphen/dot runs (the leading `5` is preceded by `1`; the trailing `9765` is followed by `2`), so the DOI is correctly ignored.

**Fix:** port the same guards to the `.sh`. POSIX `grep -E` has no lookarounds, so either use `grep -P` (PCRE):

```sh
grep -rP '(?<![\d.-])\(?\d{3}\)?[-.\s]?\d{3}[-.\s]?\d{4}(?![\d-])' "$DIST/" --include='*.html'
```

or a portable word-boundary equivalent if `-P` can't be assumed across platforms.

[...]

PR #374, opened 2026-06-21
fix(pre-deploy): stop false positives in phone PII + third-party script scans (#362, #365)

### LINKED ISSUE #362: Pre-deploy scan false positives: phone regex matches digit-runs in URLs/IDs/coords, and first-party scripts flagged as third-party
## Summary

The scaffold-shipped pre-deploy scanner (`template/scripts/pre-deploy-check.ts`) produces false positives that block deploys on legitimate content. Two distinct bugs, both still present on `main`:

1. The **phone-number PII regex** matches any 10-digit run, including non-phone numbers embedded in URLs, IDs, and coordinates.
2. The **third-party script scan** has no concept of *same-origin = first-party*, so a site's own hand-authored scripts (e.g. `public/js/*` served from `/js/...`) are flagged as unauthorized third-party scripts.

Found while deploying a content-heavy blog post with many citation links — the scan failed on link targets that contain digit runs, even though no phone number or third-party script exists anywhere on the page.

---

## Bug 1 — Phone regex matches digit-runs inside longer numbers

`template/scripts/pre-deploy-check.ts` (line ~57):

```ts
export const phonePattern = /\(?\d{3}\)?[-.\s]?\d{3}[-.\s]?\d{4}/g;
```

With no boundary anchors, this matches a 10-digit slice of any longer number. Real-world false positives from published content:

| Content | What it actually is | Matched slice |
| --- | --- | --- |
| `s41598-025-97652-6` | Nature article ID in a citation URL | `025-97652-6`-style run |
| `/web/20120120031959/` | Wayback Machine URL timestamp | `2012012003`… |
| `37.3268981241` | decimal geo-coordinate | `37326898` digit run |

None are phone numbers; all block the deploy.

### Suggested fix

Add boundary lookarounds so the pattern won't match when embedded in a longer digit/dotted/hyphenated token, while still matching a phone number that ends a sentence:

```ts
//  (?<![\d.-])  — not preceded by a digit, dot, or hyphen (rules out IDs, URLs, coordinates)
//  (?![\d-])    — not followed by a digit or hyphen; `.` deliberately allowed so
//                 "call 408-555-1234." still matches
export const phonePattern = /(?<![\d.-])\(?\d{3}\)?[-.\s]?\d{3}[-.\s]?\d{4}(?![\d-])/g;
```

This eliminates the three false positives above without weakening detection of real phone numbers. (`PII_PHONE_ALLOW` exists, but allowlisting article IDs and coordinates as "phone numbers" is the wrong remedy — the regex should not classify them as phones in the first place.)

---

## Bug 2 — First-party scripts flagged as third-party

`template/scripts/pre-deploy-check.ts` (line ~121):

```ts
export const allowedScripts = ["cloudflareinsights", "_astro", "challenges.cloudflare.com", "cdn.polar.sh", "cdn.snipcart.com", "cdn.shopify.com", "sdks.shopifycdn.com"];
```

The allowlist is a list of substrings of known external CDNs/bundles. A site that ships its own hand-authored JavaScript in `public/js/` (served same-origin as `/js/foo.js`) trips the "unauthorized third-party script" blocker, even though a same-origin script is by definition **not** third-party.

### Suggested fix

Treat same-origin / root-relative script `src` values as first-party before consulting the external-domain allowlist. Either:

- skip any `src` that is root-relative (`src="/..."`) or has no host (same-origin), and only apply `allowedScripts` to absolute cross-origin URLs; or
- at minimum, add same-origin script directories (e.g. `/js/`) to the default allowlist.

The first option is more correct — the check's intent is "no *unexpected external* scripts," and a same-origin path can never be external.

---

## Environment

- Scaffold copied from Anglesite into a Cloudflare Pages site (Astro 5 + Markdoc).
- Bugs confirmed by inspecting `template/scripts/pre-deploy-check.ts` on `main` (lines ~57 and ~121).
- Worked around downstream by editing the copied script, but the scaffold should ship correct defaults so new sites don't hit this.

davidwkeith:
Fixes #362 and #365 — the pre-deploy gate was blocking legitimate deploys with two false positives, and the `.sh` hook (`scripts/pre-deploy-check.sh`) had drifted cruder than the `.ts` scanner (`template/scripts/pre-deploy-check.ts`). Both are now aligned and fixed.

## Bug 1 — phone PII scan matched digit runs inside URLs / IDs / coordinates

The unanchored 3-3-4 pattern matched any 10-digit slice of a longer token. Real content that blocked deploys on `pulletsforever.com`:

| Content | What it is | Matched slice |
|---|---|---|
| `…/articles/s41598-025-97652-6` | Nature DOI in a citation URL | `598-025-9765` |
| `/web/20120120031959/` | Wayback Machine timestamp | `2012012003` |
| `37.3268981241` | decimal geo-coordinate | `3268981241` |

**Fix:** boundary guards that reject a run preceded by a digit, `/`, or a decimal point, and followed by a digit or `/` — while still matching:
- a number that **ends a sentence** (`…555-123-4567.`)
- a number with a **`1-` country-code prefix** (`1-800-662-4357`)

The `.ts` scanner uses lookarounds; the `.sh` hook emulates them with POSIX ERE context classes (no `grep -P`, so it stays cross-platform / works on BSD grep).

## Bug 2 — first-party scripts flagged as third-party

The script scan flagged any `<script src>` not on the allowlist, including a site's own root-relative scripts (`/js/footnote-popup.js`). A third-party gate should only flag **external** origins.

**Fix:** only scheme-qualified (`https://`) or protocol-relative (`//`) srcs are treated as third-party; root-relative (`/…`) and relative srcs are first-party by definition. The `.sh` hook now extracts individual `<script>` tags with `grep -o` so minified one-line HTML can't defeat the exclusion filter (a secondary issue noted in #365).

## Tests

- Regression tests for all three phone false positives, plus the country-code and sentence-ending cases that must keep matching.
- Tests for first-party root-relative/relative scripts (allowed) and protocol-relative external scripts (flagged), including a mixed first+third-party case.
- Light assertions on the `.sh` source so the boundary guards and external-only filter can't silently regress.
- Verified the live `.sh` hook end-to-end: the three false positives now pass; real PII (incl. country code) and external scripts (incl. `//`) are still blocked.

`npx vitest run` — the only failures are 2 pre-existing, unrelated failures in `test/apply-edit-dispatcher.test.js` (confirmed present on a clean tree); all 103 pre-deploy scan tests pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FKuH4bGQjPJ5pZJbd7USJS)_

### DIFF
--- a/scripts/pre-deploy-check.sh
+++ b/scripts/pre-deploy-check.sh
@@ -56,17 +56,35 @@ fi
 # Trailing: not followed by a digit or '/' (a trailing '.' is allowed so a
 # number that ends a sentence still matches). POSIX ERE has no lookaround, so
 # the boundaries are matched as literal context chars.
-PHONE_HITS=$(grep -rE '(^|[^0-9/.])\(?[0-9]{3}\)?[-.[:space:]]?[0-9]{3}[-.[:space:]]?[0-9]{4}([^0-9/]|$)' "$DIST/" --include='*.html' 2>/dev/null || true)
+# Extract each phone occurrence (-o keeps the boundary context chars, which the
+# digit normalization below strips). Line-based -r output can't be normalized
+# per-number, so the allowlist comparison needs the individual matches.
+PHONE_MATCHES=$(grep -rohE '(^|[^0-9/.])\(?[0-9]{3}\)?[-.[:space:]]?[0-9]{3}[-.[:space:]]?[0-9]{4}([^0-9/]|$)' "$DIST/" --include='*.html' 2>/dev/null || true)
 
-# Filter out allowlisted phone numbers (normalize to digits-only for comparison)
-if [[ -n "$PHONE_HITS" && -n "$PHONE_ALLOW" ]]; then
+# Filter out allowlisted phone numbers. Normalize BOTH the allowlist entry and
+# each hit to their last 10 digits before comparing, so formatting differences
+# (dashes vs dots, a 1- country-code prefix) never defeat the allowlist. The
+# previous digits-only grep -v matched the normalized allowlist string against
+# the still-formatted hit and so never suppressed anything (issues #362, #365).
+if [[ -n "$PHONE_MATCHES" && -n "$PHONE_ALLOW" ]]; then
+  ALLOWED_NORM=""
   IFS=',' read -ra ALLOWED_PHONES <<< "$PHONE_ALLOW"
   for phone in "${ALLOWED_PHONES[@]}"; do
-    digits=$(echo "$phone" | tr -cd '0-9')
-    if [[ -n "$digits" ]]; then
-      PHONE_HITS=$(echo "$PHONE_HITS" | grep -v "$digits" || true)
-    fi
+    d=$(printf '%s' "$phone" | tr -cd '0-9')
+    d=${d: -10}
+    [[ -n "$d" ]] && ALLOWED_NORM+="${d}"$'\n'
   done
+  PHONE_HITS=""
+  while IFS= read -r hit; do
+    [[ -z "$hit" ]] && continue
+    hd=$(printf '%s' "$hit" | tr -cd '0-9')
+    hd=${hd: -10}
+    if ! printf '%s' "$ALLOWED_NORM" | grep -qxF "$hd"; then
+      PHONE_HITS+="${hit}"$'\n'
+    fi
+  done <<< "$PHONE_MATCHES"
+else
+  PHONE_HITS="$PHONE_MATCHES"
 fi
 
 if [[ -n "$PHONE_HITS" ]]; then
--- a/scripts/pre-deploy-check.sh
+++ b/scripts/pre-deploy-check.sh
@@ -49,7 +49,14 @@ if [[ -n "$EMAIL_HITS" ]]; then
   REASONS+=("Possible email address found in built HTML")
 fi
 
-PHONE_HITS=$(grep -rE '\(?\d{3}\)?[-.[[:space:]]]?\d{3}[-.[[:space:]]]?\d{4}' "$DIST/" --include='*.html' 2>/dev/null || true)
+# Boundary guards keep the 3-3-4 shape from matching digit runs embedded in
+# longer tokens — DOIs in citation URLs, Wayback timestamps, decimal
+# coordinates (issues #362, #365). Leading: not preceded by a digit, '/', or
+# '.' (a preceding '-' is allowed so a '1-800-…' country code still matches).
+# Trailing: not followed by a digit or '/' (a trailing '.' is allowed so a
+# number that ends a sentence still matches). POSIX ERE has no lookaround, so
+# the boundaries are matched as literal context chars.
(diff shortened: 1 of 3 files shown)


davidwkeith (commented):
## Review — looks solid, two follow-up notes

I checked out the branch, ran the suite (85/85 pass), and exercised both `.sh` grep patterns and the `.ts` regexes against the real false positives. Both bugs are fixed and the `.sh` hook is now aligned with the `.ts` scanner.

**Verified working:**
- Phone scan: the Nature DOI, Wayback timestamp, and decimal coordinate no longer match; a sentence-ending number (`555-123-4567.`) and a country-code-prefixed number (`1-800-662-4357`) still match. ✅
- Script scan: root-relative (`/js/…`, `/_astro/…`) and relative srcs are ignored; only scheme-qualified and protocol-relative (`//…`) srcs are flagged. The `-o` extraction correctly handles the minified one-line case where first- and third-party tags share a line — that secondary issue from #365 is fixed too. ✅
- Cross-platform: POSIX ERE context classes, no `grep -P`. ✅
- Dropping the `_astro` exclusion in the `.sh` is safe — root-relative srcs are no longer matched at all, so there's nothing left to exclude.

### 1. (non-blocking) `.sh` phone allowlist is ineffective for formatted numbers

Out of scope for this diff (lines 61–70 are untouched), but directly adjacent to what you're fixing, so flagging it: the `.sh` allowlist normalizes the *allowlist entry* to digits-only but then `grep -v`s that digit string against the **raw, still-formatted** hit line:

```
PII_PHONE_ALLOW=800-662-4357   → digits "8006624357"
hit line: "<p>Call our hotline 800-662-4357 today.</p>"
grep -v "8006624357"           → no match → number NOT suppressed
```

So an owner who legitimately publishes a business line and allowlists it (exactly as `template/CLAUDE.md` instructs) would still have the `.sh` gate block their deploy — the same false-positive family this PR is closing. The `.ts` `scanPhones` handles it correctly because it normalizes *both* sides to the last 10 digits. Worth a follow-up to bring the `.sh` allowlist comparison in line (normalize the hit's digits before comparing).

### 2. (nit) `scriptSrcPattern` export is now dead code

`template/scripts/pre-deploy-check.ts:120` still exports `scriptSrcPattern`, but after this refactor `scanScripts` builds its own local regex and nothing else references the export (the only other `scriptSrcPattern` is an unrelated local in `perf-budget.ts`). Safe to delete.

Neither is a blocker — the PR does what it says and the test coverage is good.

---

davidwkeith:
Both notes addressed in 624732c:

1. **`.sh` phone allowlist** — fixed. The allowlist filter now normalizes *both* the allowlist entry and each extracted hit to their last 10 digits before comparison (matching `.ts` `scanPhones`), instead of comparing a normalized allowlist string against a still-formatted hit. Verified live: `PII_PHONE_ALLOW=1-800-662-4357` now suppresses a published `800-662-4357`, while a second, non-allowlisted number still blocks. Folded in here rather than as a follow-up since it's the same false-positive family this branch closes.

2. **Dead `scriptSrcPattern` export** — removed. Confirmed nothing else references it (the `perf-budget.ts` occurrence is an unrelated local); `perf-budget` tests still pass.

Added a regression assertion that the `.sh` allowlist comparison normalizes (no more `grep -v "$digits"`). Full pre-deploy scan suite green (86 + 18 tests).

---

PR #375, opened 2026-06-22
refactor(csp): decouple central CSP builder from optional provider modules (#364)

### LINKED ISSUE #364: Pre-deploy scan vendors ecommerce/booking CSP modules unconditionally onto content-only sites
## Summary

The 1.1.0 pre-deploy security/SEO scan vendors five commerce provider helper modules into every consuming site's `scripts/` directory — `snipcart.ts`, `shopify-buy-button.ts`, `paddle.ts`, `lemon-squeezy.ts`, and `booking.ts` (~508 lines combined). They ship and are imported unconditionally, even on sites that have no checkout, booking, or product catalog.

## Why this is a problem

`scripts/csp.ts` does a static top-level import of each provider's `build*CSP()` helper:

```ts
import { buildSnipcartCSP } from "./snipcart.js";
import { buildShopifyCSP } from "./shopify-buy-button.js";
import { buildPaddleCSP } from "./paddle.js";
import { buildLemonSqueezyCSP } from "./lemon-squeezy.js";
import { buildBookingCSP, type BookingProvider } from "./booking.js";
```

So the files can't simply be deleted by a site owner without also editing `csp.ts` — they're load-bearing for the module graph even when `ECOMMERCE_PROVIDER` / `BOOKING_PROVIDER` are unset. The result is unavoidable dead code on content-only sites (blogs, docs, personal sites): the helpers are never reachable at runtime, but they sit in source, get type-checked, and show up in every audit.

## Impact

- Personal/blog sites that adopt the 1.1.0 scan carry ecommerce + booking code they will never run.
- Removing it is a manual fork of `csp.ts` (narrow the `ecommerce` union, drop the `booking` field, delete the import lines and the provider branches), which then drifts from upstream on the next update.

## Suggested fixes (any one)

1. **Lazy-load provider modules** — only `await import("./snipcart.js")` inside the matching `if (providers.ecommerce === "snipcart")` branch, so unused providers aren't part of the static graph and the files can be removed without touching `csp.ts`.
2. **Gate vendoring on enabled providers** — have the scaffolding/update step copy a provider's helper only when its `.site-config` key is set, instead of vendoring all of them by default.
3. **Move provider CSP builders into the relevant skill** (`buy-button`, `snipcart`, `booking`, etc.) so they land only when that skill is actually installed.

## Context

Found while re-platforming https://pulletsforever.com onto Anglesite 1.1.0 (a personal blog — no commerce). I removed the five files and the `csp.ts` references downstream; filing here so the upstream scan doesn't force them on content-only sites.

davidwkeith:
Fixes #364.

## Problem

`template/scripts/csp.ts` statically imported five commerce/booking provider helper modules just to call their `build*CSP()` helpers:

```ts
import { buildSnipcartCSP } from "./snipcart.js";
import { buildShopifyCSP } from "./shopify-buy-button.js";
import { buildPaddleCSP } from "./paddle.js";
import { buildLemonSqueezyCSP } from "./lemon-squeezy.js";
import { buildBookingCSP, type BookingProvider } from "./booking.js";
```

That made all five modules **load-bearing for the module graph** even on content-only sites (blogs, docs, personal sites) with no checkout or booking. A site owner couldn't delete the dead helpers without also hand-editing `csp.ts` — a manual fork that drifts from upstream on every update.

## Fix

Centralize the five CSP builders into `csp.ts`, alongside the ones that **already** live there inline (`buildPolarCSP`, `buildTurnstileCSP`, `buildGiscusCSP`, `buildPodcastYouTubeCSP`, `buildTrackingCSP`), and drop the static imports. `BookingProvider` is now defined locally in `csp.ts` so there's no type dependency on `booking.ts` either.

- **CSP output is byte-for-byte unchanged** — `buildCSP`'s branches call the same-named builders, now local.
- **`buildCSP` stays synchronous** — chose centralization over the dynamic-import option in the issue specifically to avoid an `async` ripple across its ~30 call sites and the skill snippets.
- Nothing in the site build graph imports the provider helper modules anymore, so `snipcart.ts` / `shopify-buy-button.ts` / `paddle.ts` / `lemon-squeezy.ts` / `booking.ts` can be deleted on a content-only site **without touching `csp.ts`** — exactly the outcome the issue asked for.

## Changes

- Removed `build*CSP` + their per-provider CSP interfaces from the five provider modules. `booking.ts` keeps the `BookingProvider` type (still used by its embed helpers).
- Moved the builder unit tests into `csp.test.ts`; added a **guard test** asserting `csp.ts` never re-imports a provider module (prevents regression of #364).
- Repointed `skills/booking/SKILL.md` to `csp.ts` for `buildBookingCSP()`.

## Tests

Full suite green except the 2 pre-existing, unrelated `test/apply-edit-dispatcher.test.js` failures (present on `main`). The CSP + all five provider suites pass (86 + 196 across the touched files).

### Note on default vendoring

This resolves the issue's core grievance (the modules are no longer load-bearing / no `csp.ts` fork needed to remove them). The issue also floated *gating vendoring* so the files aren't scaffolded at all unless a provider is configured — that's a separate `scaffold.sh`/`update.sh` change and can be a follow-up if desired.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FKuH4bGQjPJ5pZJbd7USJS)_

### DIFF
--- a/skills/booking/SKILL.md
+++ b/skills/booking/SKILL.md
@@ -153,7 +153,7 @@ For floating style (site-wide), add the structured data to the root layout or ho
 
 ### Content Security Policy
 
-Read `public/_headers` and update the CSP to allow the provider's domains. Use `buildBookingCSP()` from `${CLAUDE_PLUGIN_ROOT}/template/scripts/booking.ts` to get the required domains:
+Read `public/_headers` and update the CSP to allow the provider's domains. Use `buildBookingCSP()` from `${CLAUDE_PLUGIN_ROOT}/template/scripts/csp.ts` to get the required domains:
 
 **Cal.com:** Add `app.cal.com` to `script-src`, `style-src`, and `frame-src`.
 
--- a/template/scripts/booking.ts
+++ b/template/scripts/booking.ts
@@ -194,34 +194,3 @@ export function buildCalendlyEmbed(
   // button → popup
   return `${cssLink}\n<button class="booking-button" onclick="Calendly.initPopupWidget({url:'${url}'});return false;">${safeButtonText}</button>\n${jsScript}`;
 }
-
-// ---------------------------------------------------------------------------
-// CSP directives
-// ---------------------------------------------------------------------------
-
-/** CSP directives needed for a booking provider */
-export interface BookingCSP {
-  "script-src": string[];
-  "style-src": string[];
-  "frame-src": string[];
-}
-
-/**
- * Build CSP directives required for a booking provider.
- * @param provider - "cal" or "calendly"
- * @returns Object with arrays of domains to add to each CSP directive
- */
-export function buildBookingCSP(provider: BookingProvider): BookingCSP {
-  if (provider === "cal") {
-    return {
-      "script-src": ["app.cal.com"],
-      "style-src": ["app.cal.com"],
-      "frame-src": ["app.cal.com"],
-    };
-  }
-  return {
-    "script-src": ["assets.calendly.com"],
-    "style-src": ["assets.calendly.com"],
-    "frame-src": ["calendly.com"],
-  };
-}
--- a/template/scripts/csp.ts
+++ b/template/scripts/csp.ts
@@ -8,16 +8,19 @@
  * @module
  */
 
-import { buildSnipcartCSP } from "./snipcart.js";
-import { buildShopifyCSP } from "./shopify-buy-button.js";
-import { buildPaddleCSP } from "./paddle.js";
-import { buildLemonSqueezyCSP } from "./lemon-squeezy.js";
-import { buildBookingCSP, type BookingProvider } from "./booking.js";
 import { readConfigFromString } from "./config.js";
 
 // ---------------------------------------------------------------------------
 // Types
 // ---------------------------------------------------------------------------
+
+/**
+ * Supported booking providers. Defined here (rather than imported from
+ * `booking.ts`) so this central CSP module has no runtime or type dependency
+ * on the optional provider helper modules — content-only sites can delete
+ * `booking.ts`, `snipcart.ts`, etc. without having to edit this file (#364).
+ */
+export type BookingProvider = "cal" | "calendly";
 
 /** Parsed provider configuration from .site-config */
 export interface SiteProviders {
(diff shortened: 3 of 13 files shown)


davidwkeith (commented):
**Review (LGTM — posted as comment since GitHub won't let me approve my own PR)**

Clean, well-scoped refactor that correctly resolves #364. Traced the full integration:

**Correctness**
- Decoupling goal met: after this change `csp.ts` has zero runtime *or* type dependency on the five provider modules, so a content-only site can delete `snipcart.ts` / `shopify-buy-button.ts` / `paddle.ts` / `lemon-squeezy.ts` / `booking.ts` without touching `csp.ts`. The new source-reading guard test locks this against regression.
- "Byte-for-byte unchanged" output checks out. The one subtlety — old `buildShopifyCSP` returned `"frame-src": []`, new one omits it — is equivalent, since `buildCSP` only emits `frame-src` when `length > 0`. Same header string.
- Staying synchronous (centralization over the issue's `await import()` option) was the right call — avoids making `buildCSP` async across its ~30 call sites. The unified `CSPDirectives` return type is a superset of the old per-provider interfaces, so nothing is lost.
- No stale references: `skills/booking/SKILL.md` is repointed to `csp.ts`; `booking.ts` correctly retains `BookingProvider`/`BookingStyle` for its embed helpers.

**Tests & CI**
- Builder unit tests relocated into `csp.test.ts` (removed from provider test files) rather than duplicated — no orphaned imports. Per-builder coverage preserved.
- Both CI checks green on head. The 2 `apply-edit-dispatcher.test.js` failures noted in the description are pre-existing on `main` and unrelated.

**Minor (non-blocking)**
- Two structurally-identical `BookingProvider` definitions now exist (`csp.ts` + `booking.ts`) — intended decoupling cost, interchangeable under structural typing.
- `CSPDirectives` stays a non-exported interface used as an exported function's return type; only matters under `.d.ts` emit (TS4053), which isn't enabled here, and `buildPolarCSP` already did this.

Deferring the "gate vendoring at scaffold time" idea (issue option #2) to a follow-up is reasonable. LGTM.

---

PR #376, opened 2026-06-22
fix(indieweb): harden Step 1 gate to verify @dwk/* packages import, not just exist (#363)

davidwkeith:
## What

Hardens the `skills/indieweb/SKILL.md` **Step 1 — Check package availability** gate so it verifies the `@dwk/*` packages actually **import under Node ESM**, not merely that they're published.

## Why (issue #363, note 1)

The previous gate ran only `npm view <pkg> version`, which confirms existence. The `0.1.0-beta.*` `@dwk/*` builds pass that check but throw `ERR_MODULE_NOT_FOUND` on import — their compiled `dist/` re-exports omit file extensions, which Node ESM can't resolve. Composing such a package into `site-entry.js` produces a Worker that can't boot, and the breakage stays invisible because the repo's tests alias `@dwk/*` to stubs (`tests/__stubs__/dwk-*.ts`) and wrangler bundling can paper over it at build time. Net effect: a fresh `/anglesite:indieweb` run today would wave broken packages straight through the gate.

## What changed

- Step 1 now installs the four packages (`@dwk/indieauth`, `@dwk/webmention`, `@dwk/micropub`, `@dwk/webauthn`) into a throwaway `mktemp` directory — never touching the site — and dynamically imports each one, **failing closed** on any install or import error. A published-but-broken build no longer passes; the owner gets the existing "not available yet" message.
- `allowed-tools` gains `Bash(node *)`, `Bash(mktemp *)`, `Bash(printf *)`, `Bash(rm *)` so the smoke test runs without prompting the (often non-technical) owner.

## Verification

- Gate **fails closed** (exit 1 + `FAIL @dwk/indieauth: ERR_MODULE_NOT_FOUND`) against the current published `0.1.0-beta.2` packages.
- Gate **passes** (exit 0, `ok`) for packages that import cleanly.
- `tests/build-instructions.test.ts` — 21/21 green (SKILL.md still validates).
- Full suite: only 2 pre-existing, unrelated `test/apply-edit-dispatcher.test.js` failures (a root-in-container filesystem-permission quirk; fail identically on clean `HEAD`).

## Scope note

This addresses the **in-repo** half of #363 note 1. The upstream root cause — the `@dwk/*` `dist/` re-exports needing `.js` extensions — lives in the separate `@dwk/*` package repo and is still unfixed (verified: `0.1.0-beta.2` still fails to import). Issue #363 notes 2 (API realignment) and 3 (webmention render XSS) were already fixed by #370 and #369 respectively.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QTFyxU3zyrr3QauRU2Tmkv)_

### DIFF
--- a/skills/indieweb/SKILL.md
+++ b/skills/indieweb/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: indieweb
 description: "Deploy self-owned IndieAuth, Webmention, and Micropub endpoints on your domain"
-allowed-tools: Bash(npm run build), Bash(npm install *), Bash(npm view *), Bash(npx wrangler *), Bash(openssl *), Bash(grep *), Bash(gh *), Write, Read, Edit, Glob, mcp__cloudflare__accounts_list, mcp__cloudflare__set_active_account, mcp__cloudflare__d1_database_create, mcp__cloudflare__d1_databases_list, mcp__cloudflare__d1_database_get, mcp__cloudflare__r2_bucket_create, mcp__cloudflare__r2_buckets_list
+allowed-tools: Bash(npm run build), Bash(npm install *), Bash(npm view *), Bash(npx wrangler *), Bash(openssl *), Bash(grep *), Bash(gh *), Bash(node *), Bash(mktemp *), Bash(printf *), Bash(rm *), Write, Read, Edit, Glob, mcp__cloudflare__accounts_list, mcp__cloudflare__set_active_account, mcp__cloudflare__d1_database_create, mcp__cloudflare__d1_databases_list, mcp__cloudflare__d1_database_get, mcp__cloudflare__r2_bucket_create, mcp__cloudflare__r2_buckets_list
 disable-model-invocation: true
 ---
 
@@ -33,25 +33,28 @@ If `INDIEWEB_ENABLED=true` is already set in `.site-config`, the endpoints are a
 
 ## Step 1 — Check package availability
 
-The `@dwk/*` packages must be published on npm before the endpoints can be installed. Check each one:
+The `@dwk/*` packages must be published on npm **and actually load under Node ESM** before the endpoints can be installed. A published version is not enough: a broken build whose compiled `dist/` re-exports omit file extensions passes `npm view` but throws `ERR_MODULE_NOT_FOUND` the moment it's imported (issue #363). Composing such a package into `site-entry.js` produces a Worker that can't boot — and the breakage is invisible until deploy because the repo's tests alias `@dwk/*` to stubs. So the gate must **import** each package, not just confirm it exists.
+
+Run this combined gate. It installs the four packages into a throwaway directory (never touching the site) and dynamically imports each one, failing closed on any error:
 
 ```sh
-npm view @dwk/indieauth version
+PKGS="@dwk/indieauth @dwk/webmention @dwk/micropub @dwk/webauthn"
+SMOKE=$(mktemp -d) && printf '{"type":"module","private":true}' > "$SMOKE/package.json"
+( cd "$SMOKE" \
+  && npm install $PKGS >/dev/null 2>&1 \
+  && for p in $PKGS; do \
+       node -e "import('$p').then(()=>console.log('ok   '+process.argv[1])).catch(e=>{console.error('FAIL '+process.argv[1]+': '+(e.code||e.message));process.exit(1)})" "$p" || exit 1; \
+     done )
+GATE=$?
+rm -rf "$SMOKE"
+echo "gate exit: $GATE"
 ```
 
-```sh
-npm view @dwk/webmention version
-```
+If `gate exit` is non-zero — any package is unpublished, reports version `0.0.0`, fails to install, or prints a `FAIL` line (does not import) — stop and tell the owner:
 
-```sh
-npm view @dwk/micropub version
-```
+"The IndieWeb endpoint packages aren't ready yet — this feature isn't available right now. It's being built by an open-source project and will be ready soon. I'll let you know when it ships. In the meantime, your site already supports the passive IndieWeb (microformats, `rel="me"`, RSS) — see `docs/indieweb.md` for what's already working."
 
-If **any** package returns an error (not found) or reports version `0.0.0`, stop and tell the owner:
-
-"The IndieWeb endpoint packages aren't published yet — this feature isn't available right now. It's being built by an open-source project and will be ready soon. I'll let you know when it ships. In the meantime, your site already supports the passive IndieWeb (microformats, `rel="me"`, RSS) — see `docs/indieweb.md` for what's already working."
-
-Do not attempt git installs, forks, or workarounds. The gate is intentional.
+Do not attempt git installs, forks, version pins to a broken build, or workarounds. The gate is intentional: a package that imports cleanly is the contract the rest of this skill depends on.
 
 ## Step 2 — Choose endpoints
 
--- a/skills/indieweb/SKILL.md
+++ b/skills/indieweb/SKILL.md
@@ -5,7 +5,7 @@ allowed-tools: Bash(npm run build), Bash(npm install *), Bash(npx wrangler *), B
 disable-model-invocation: true
 ---
 
-Set up self-owned IndieWeb endpoints on the owner's primary domain using the `@dwk/*` worker packages. Deploys IndieAuth (sign in with your domain), Webmention (receive cross-site mentions), and Micropub (publish from third-party clients) — each independently gated and composed into the site's existing `site-entry.js` Worker.
+Set up self-owned IndieWeb endpoints on the owner's primary domain using the `@dwk/*` worker packages. Deploys IndieAuth (sign in with your domain), Webmention (receive cross-site mentions), and Micropub (publish from third-party clients) — each independently selectable, runtime-guarded by its own bindings, and composed into the site's existing `site-entry.js` Worker.
 
 The owner's identity, tokens, and data stay on their own Cloudflare account. No third-party identity or mention hosts.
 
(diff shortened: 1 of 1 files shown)


skills/indieweb/SKILL.md
```diff
@@ -33,25 +33,28 @@ If `INDIEWEB_ENABLED=true` is already set in `.site-config`, the endpoints are a
...
+Run this combined gate. It installs the four packages into a throwaway directory (never touching the site) and dynamically imports each one, failing closed on any error:
 
 ```sh
-npm view @dwk/micropub version
+PKGS="@dwk/indieauth @dwk/webmention @dwk/micropub @dwk/webauthn"
+SMOKE=$(mktemp -d) && printf '{"type":"module","private":true}' > "$SMOKE/package.json"
+( cd "$SMOKE" \
+  && npm install $PKGS >/dev/null 2>&1 \
+  && for p in $PKGS; do \
+       node -e "import('$p').then(()=>console.log('ok   '+process.argv[1])).catch(e=>{console.error('FAIL '+process.argv[1]+': '+(e.code||e.message));process.exit(1)})" "$p" || exit 1; \
```
davidwkeith: **Bug: `process.argv[1]` should be `process.argv[2]`**

When Node is invoked with `-e "code" "$p"`, the positional arguments are:

| index | value |
|---|---|
| `process.argv[0]` | path to the `node` executable |
| `process.argv[1]` | `''` (empty string) — no script file when using `-e` |
| `process.argv[2]` | `$p` (the package name passed as the trailing argument) |

So `process.argv[1]` in both the `ok` and `FAIL` log lines will always be an empty string, not the package name. The gate exit code is still correct, but every diagnostic line prints as `ok   ` or `FAIL : ERR_MODULE_NOT_FOUND` with no package name visible — making it impossible to tell which package failed without re-running manually.

Fix:
```diff
-       node -e "import('$p').then(()=>console.log('ok   '+process.argv[1])).catch(e=>{console.error('FAIL '+process.argv[1]+': '+(e.code||e.message));process.exit(1)})" "$p" || exit 1; \
+       node -e "import('$p').then(()=>console.log('ok   '+process.argv[2])).catch(e=>{console.error('FAIL '+process.argv[2]+': '+(e.code||e.message));process.exit(1)})" "$p" || exit 1; \
```

---

skills/indieweb/SKILL.md
```diff
@@ -1,7 +1,7 @@
 ---
 name: indieweb
 description: "Deploy self-owned IndieAuth, Webmention, and Micropub endpoints on your domain"
-allowed-tools: Bash(npm run build), Bash(npm install *), Bash(npm view *), Bash(npx wrangler *), Bash(openssl *), Bash(grep *), Bash(gh *), Write, Read, Edit, Glob, mcp__cloudflare__accounts_list, mcp__cloudflare__set_active_account, mcp__cloudflare__d1_database_create, mcp__cloudflare__d1_databases_list, mcp__cloudflare__d1_database_get, mcp__cloudflare__r2_bucket_create, mcp__cloudflare__r2_buckets_list
+allowed-tools: Bash(npm run build), Bash(npm install *), Bash(npm view *), Bash(npx wrangler *), Bash(openssl *), Bash(grep *), Bash(gh *), Bash(node *), Bash(mktemp *), Bash(printf *), Bash(rm *), Write, Read, Edit, Glob, mcp__cloudflare__accounts_list, mcp__cloudflare__set_active_account, mcp__cloudflare__d1_database_create, mcp__cloudflare__d1_databases_list, mcp__cloudflare__d1_database_get, mcp__cloudflare__r2_bucket_create, mcp__cloudflare__r2_buckets_list
```
davidwkeith: **Dead permission: `Bash(npm view *)` is no longer used**

The old Step 1 ran `npm view @dwk/* version` — those calls are gone in this PR. No other part of the skill uses `npm view`. The permission should be dropped to keep `allowed-tools` minimal.

```diff
-allowed-tools: Bash(npm run build), Bash(npm install *), Bash(npm view *), Bash(npx wrangler *), ...
+allowed-tools: Bash(npm run build), Bash(npm install *), Bash(npx wrangler *), ...
```

---

skills/indieweb/SKILL.md
```diff
@@ -33,25 +33,28 @@ If `INDIEWEB_ENABLED=true` is already set in `.site-config`, the endpoints are a
...
-npm view @dwk/webmention version
-```
+Run this combined gate. It installs the four packages into a throwaway directory (never touching the site) and dynamically imports each one, failing closed on any error:
 
 ```sh
-npm view @dwk/micropub version
+PKGS="@dwk/indieauth @dwk/webmention @dwk/micropub @dwk/webauthn"
+SMOKE=$(mktemp -d) && printf '{"type":"module","private":true}' > "$SMOKE/package.json"
+( cd "$SMOKE" \
+  && npm install $PKGS >/dev/null 2>&1 \
```
davidwkeith: **`npm install` errors are silenced — transient failures look identical to broken packages**

`>/dev/null 2>&1` swallows all npm output. If the install fails due to a transient network issue, a private registry timeout, or a permissions problem, the owner receives the same "packages aren't ready yet" message as an actually broken build — and is left waiting for a ship date when their real problem is a flaky connection.

Suggested fix: keep stdout suppressed but surface install errors on failure so the owner (or the model reading the output) has something actionable:

```sh
INSTALL_LOG=$(npm install $PKGS 2>&1 >/dev/null) || { echo "npm install failed:"; echo "$INSTALL_LOG"; exit 1; }
```

Or at minimum change the owner-facing message to acknowledge the network case: "The IndieWeb endpoint packages couldn't be installed — they may not be ready yet, or there may be a temporary network issue. Try again in a minute; if it keeps failing, the packages may be in an incompatible state."

---

skills/indieweb/SKILL.md
```diff
@@ -33,25 +33,28 @@ If `INDIEWEB_ENABLED=true` is already set in `.site-config`, the endpoints are a
...
+  && for p in $PKGS; do \
+       node -e "import('$p').then(()=>console.log('ok   '+process.argv[1])).catch(e=>{console.error('FAIL '+process.argv[1]+': '+(e.code||e.message));process.exit(1)})" "$p" || exit 1; \
+     done )
+GATE=$?
+rm -rf "$SMOKE"
```
davidwkeith: **`rm -rf "$SMOKE"` with a potentially empty variable**

`SMOKE=$(mktemp -d) && printf ... > "$SMOKE/package.json"` — if `mktemp -d` fails (e.g. `/tmp` is full), `$SMOKE` is empty and `rm -rf ""` runs. GNU `rm` typically errors rather than deleting the working directory in this case, but it's fragile. Guard it:

```sh
[ -n "$SMOKE" ] && rm -rf "$SMOKE"
```

---

skills/indieweb/SKILL.md
```diff
@@ -33,25 +33,28 @@ If `INDIEWEB_ENABLED=true` is already set in `.site-config`, the endpoints are a
...
-```
-
-```sh
-npm view @dwk/webmention version
-```
+Run this combined gate. It installs the four packages into a throwaway directory (never touching the site) and dynamically imports each one, failing closed on any error:
 
 ```sh
-npm view @dwk/micropub version
+PKGS="@dwk/indieauth @dwk/webmention @dwk/micropub @dwk/webauthn"
```
davidwkeith, line 41: **Gate is atomic across all 4 packages regardless of which endpoints the owner selected**

The skill's opening paragraph says the three endpoints are "each independently gated." But this gate installs and imports all four packages — including `@dwk/webauthn` and `@dwk/indieauth` — even if the owner only wants Webmention (which has no dependency on IndieAuth or WebAuthn).

If `@dwk/indieauth` has the broken `dist/` re-exports issue but `@dwk/webmention` imports cleanly, an owner who only wants Webmention is still blocked by an unrelated breakage.

Options:
1. Keep the gate atomic and document that it's all-or-nothing (simplest, but contradicts "independently gated").
2. Parameterise the gate to check only the packages for selected endpoints. Webmention needs `@dwk/webmention`; IndieAuth needs `@dwk/indieauth` + `@dwk/webauthn`; Micropub needs all four. The gate would run before the endpoint selection (Step 2) in its current position, so option 1 or running the gate after Step 2 would both work.

---

davidwkeith (commented):
The approach is sound — importing rather than `npm view`-ing is the right fix for the broken-build-passes-gate problem. One confirmed bug and three other issues worth addressing before merge.

**Must fix:**
- `process.argv[1]` → `process.argv[2]` in the `node -e` one-liner (line 46). With `-e`, Node sets `argv[1]` to `''`, so every `ok` and `FAIL` line prints with a blank package name. The gate exit code is still correct, but diagnostic output is broken.

**Should fix:**
- `Bash(npm view *)` is now dead in `allowed-tools` — the old `npm view` calls are gone and nothing else in the skill uses it (line 4).
- `npm install` stderr is fully silenced. A transient network failure produces the same "packages aren't ready" owner message as a genuinely broken build, with no way to distinguish the two (line 44).

**Worth discussing:**
- The gate checks all 4 packages atomically, but the skill describes the three endpoints as "independently gated." An owner who only wants Webmention is blocked if `@dwk/indieauth` fails, even though Webmention doesn't depend on it (line 41).

---

davidwkeith:
Thanks for the thorough review. All five points addressed across `bb4e362` and `e7573ca`:

**1. `argv[1]` → `argv[2]` (must-fix)** — Held off on the literal change, because `argv[1]` is actually correct here. On Node 22 (the project's required runtime), `node -e "code" "$pkg"` puts the trailing arg at `process.argv[1]`, not `argv[2]`:
```
$ node --version
v22.22.2
$ node -e "console.log('argv1=['+process.argv[1]+'] argv2=['+process.argv[2]+']')" "@dwk/webmention"
argv1=[@dwk/webmention] argv2=[undefined]
```
The `argv[1]=''` behavior holds for a *bare* `-e` with no trailing positionals; once you pass `"$pkg"`, it lands at `argv[1]`. So the suggested `argv[2]` would have printed `undefined`. That said — the diagnostic value was real, and the argv indexing is an easy footgun, so I dropped argv entirely and interpolate `$p` directly into the `node -e` string (it's already shell-double-quoted). Output now reads `ok   @dwk/webmention` / `FAIL @dwk/indieauth: ERR_MODULE_NOT_FOUND`.

**2. Dead `Bash(npm view *)` permission** — Removed from `allowed-tools`. ✅

**3. Silenced `npm install` errors** — Now captured and surfaced. The gate prints a distinct `INSTALL FAILED` block (with the npm error) for install failures vs a `FAIL <pkg>` line for broken imports, and the skill gives the owner a network-aware message for the install case so a flaky connection isn't mistaken for an unshipped package. ✅

**4. `rm -rf "$SMOKE"` empty-var guard** — The whole gate now runs inside `if [ -n "$SMOKE" ]`, so a failed `mktemp` fails closed (`GATE=1`) and never reaches the `rm`. ✅

**5. Atomic vs per-endpoint gating** — Kept atomic, reconciled the wording (`e7573ca`). The four `@dwk/*` packages ship from one monorepo at a shared version, so they break and recover together; gating them as a set keeps Step 1 simple. Reworded the "independently gated" intro to make clear the independence is about endpoint *selection* (Step 2) and runtime binding guards, not the availability check, and added a note at the gate explaining the all-or-nothing choice.

Verified all three gate paths after the rewrite: fails closed on the broken beta.2 packages (`FAIL` line now names the package), passes for importable packages, and an install error prints the distinct `INSTALL FAILED` block. `tests/build-instructions.test.ts` stays green (21/21).

---

PR #377, opened 2026-06-22
fix(indieweb): realign Micropub to the real @dwk/micropub API + reconcile TOKEN_SIGNING_KEY

davidwkeith:
## What & why

Completes the **Micropub realignment (#2c)** that the passkey IndieAuth PR (#371) explicitly deferred as out-of-scope, plus the **IndieAuth/Micropub signing-key reconciliation** — both called out in the follow-up on #363.

Investigation against the published `@dwk/micropub@0.1.0-beta.2` dist turned up that the template's Micropub wiring couldn't run against the real package — and the test stub was masking every gap (it still exported the old `createHandler`). Three layers were wrong:

### 1. `worker/site-entry.js` — Micropub handler
- **Import:** `createHandler as createMicropub` → the package exports **`createMicropub`**. The old alias resolved to `undefined` and threw at module load.
- **Per-env construction:** built once at module load with a bare `{ generatePostUrl }`, so the real `resolveConfig` threw *"`baseUrl` is not a valid URL"* on boot. Now memoized per-env (like `authFor`/`webmentionFor`) with the real `{ baseUrl, me, generatePostUrl }` derived from `SITE_URL`.
- **`generatePostUrl` signature:** the package calls it `(post, commands)`, not `(slug)`. Fixed, still publishing to `/notes/<slug>/` so the bridge path is unchanged.

### 2. `worker/indieweb-bridge.js` — D1→GitHub bridge
The package's `posts` table is keyed by **`url`** (columns `url, type, properties, deleted, created_at, updated_at`; timestamps in **Unix seconds**) and has **no sync state**. The bridge was querying `id`, `slug`, and `WHERE synced = 0` — none of which exist, so it errored on every run (same class of mismatch as the webmention inbox in #372).

- Realigned the query to the real schema.
- The bridge now owns a side **`bridge_sync(url, synced_at)`** table (mirrors the webmention custom-store precedent) and re-commits a post whenever its `updated_at` advances past the last sync — covering create / edit / delete / undelete.
- Slug derived from the post url; Unix-seconds timestamps converted to ISO.

### 3. Signing key — `TOKEN_SIGNING_KEY`
Both `@dwk/indieauth` and `@dwk/micropub` read **`env.TOKEN_SIGNING_KEY`** (one signs access tokens, the other verifies), but the skill provisioned `INDIEAUTH_SIGNING_KEY` — so the endpoints would throw *"missing required secret binding `TOKEN_SIGNING_KEY`"*.

- Renamed in the skill, the deploy-scan patterns (`pre-deploy-check.ts` + `.sh`, **keeping the legacy name matched** so pre-rename sites are still caught), and the docs.
- The skill now sets the `SITE_URL` var for **any** selected endpoint, not just Webmention — IndieAuth, WebAuthn, and Micropub all derive their origin from it.

## Tests
- Migrated the `dwk-micropub` stub to `createMicropub`.
- Rewrote the bridge sync tests for the new schema + `bridge_sync` bookkeeping; added `slugFromUrl` and `TOKEN_SIGNING_KEY` scan coverage.
- Full suite: only 2 failures, both in `test/apply-edit-dispatcher.test.js` and **pre-existing on `main`** (unrelated to this change). All IndieWeb/Micropub/bridge/scan suites pass.

## Notes for the reviewer
- The bridge sync-state approach (a bridge-owned side table vs. ALTERing the package's table) follows the webmention custom-store decision from #372 — flagging it explicitly in case you'd prefer a different mechanism.
- I noticed a **pre-existing** `readCookie` redeclaration in `site-entry.js` (an `import` + a local `function`, present since #371). esbuild/wrangler tolerate it and it's orthogonal to this work, so I left it untouched — happy to file/fix separately.
- Finding 1 from #363 (the `@dwk/*` betas not loading under plain Node ESM) remains an upstream package fix and is out of scope here.

Refs #363

https://claude.ai/code/session_01WhHDFxRw8eZdbxTPMyc5vS

---
_Generated by [Claude Code](https://claude.ai/code/session_01WhHDFxRw8eZdbxTPMyc5vS)_


scripts/pre-deploy-check.sh
```diff
@@ -106,13 +106,14 @@ if grep -rE 'pat[A-Za-z0-9]{14}\.[A-Za-z0-9]{32,}|sk-[A-Za-z0-9]{20,}|gh[pousr]_
...
+# INDIEWEB_REG_TOKEN, GITHUB_TOKEN) committed as literals. INDIEAUTH_SIGNING_KEY
+# stays matched for sites scaffolded before the TOKEN_SIGNING_KEY rename.
+# Name-only references never match — env.GITHUB_TOKEN, ${{ secrets.GITHUB_TOKEN }},
+# and `wrangler secret put TOKEN_SIGNING_KEY` all lack a credential-shaped value
 # after = or :
-if grep -rE '(INDIEAUTH_SIGNING_KEY|INDIEAUTH_SESSION_KEY|INDIEWEB_REG_TOKEN|GITHUB_TOKEN)["'\'']?[[:space:]]*[:=][[:space:]]*["'\'']?[A-Za-z0-9+/_-]{16,}' "${TOKEN_SCAN_PATHS[@]}" 2>/dev/null | grep -q .; then
-  REASONS+=("An IndieWeb secret (INDIEAUTH_SIGNING_KEY / INDIEAUTH_SESSION_KEY / INDIEWEB_REG_TOKEN / GITHUB_TOKEN) is committed in source — rotate it and store it with 'wrangler secret put'")
+if grep -rE '(TOKEN_SIGNING_KEY|INDIEAUTH_SIGNING_KEY|INDIEAUTH_SESSION_KEY|INDIEWEB_REG_TOKEN|GITHUB_TOKEN)["'\'']?[[:space:]]*[:=][[:space:]]*["'\'']?[A-Za-z0-9+/_-]{16,}' "${TOKEN_SCAN_PATHS[@]}" 2>/dev/null | grep -q .; then
+  REASONS+=("An IndieWeb secret (TOKEN_SIGNING_KEY / INDIEAUTH_SESSION_KEY / INDIEWEB_REG_TOKEN / GITHUB_TOKEN) is committed in source — rotate it and store it with 'wrangler secret put'")
 fi
```
davidwkeith, line 117: The grep pattern on the line above now includes `INDIEAUTH_SIGNING_KEY` (for legacy sites), but this REASONS message no longer lists it. A site that committed the old key name will trigger the scan but get an error message that doesn't name the secret that fired it, which makes it hard to debug. Add `INDIEAUTH_SIGNING_KEY` back to the parenthetical:

```sh
REASONS+=("An IndieWeb secret (TOKEN_SIGNING_KEY / INDIEAUTH_SIGNING_KEY (legacy) / INDIEAUTH_SESSION_KEY / INDIEWEB_REG_TOKEN / GITHUB_TOKEN) is committed in source — rotate it and store it with 'wrangler secret put'")
```

---

template/worker/indieweb-bridge.js
```diff
@@ -34,42 +49,86 @@ export async function sync(env) {
...
     .bind(BATCH_SIZE)
     .all();
 
   if (!rows.results || rows.results.length === 0) return;
 
   for (const row of rows.results) {
+    const slug = slugFromUrl(row.url);
+    if (!slug) {
+      console.error(`Bridge sync: cannot derive slug from url ${row.url}`);
+      continue;
```
davidwkeith, line 84: `continue` here leaves the row out of `bridge_sync`, so the unsynced query (`b.synced_at IS NULL`) picks it up again on every subsequent cron run. A URL that doesn't match the `/notes/<slug>/` shape will spam the Worker logs indefinitely.

Two options:
- Record a sentinel in `bridge_sync` (e.g. `synced_at = -1`) so the row is excluded from future runs, and log once.
- Or accept it as permanent — but document that the cron will retry it forever and rate-limit the log with a counter.

The sentinel approach is simpler and mirrors the existing error-path design (failed GitHub commits also skip the `bridge_sync` insert, but those are transient failures where retry makes sense — a bad URL slug is not transient).

---

template/worker/site-entry.js
```diff
@@ -229,6 +254,28 @@ function authFor(env) {
...
 }
 
+// @dwk/micropub also needs request-time config: `baseUrl`/`me` come from
+// env.SITE_URL, so the handler is built per-env (not at module load) and
+// memoized. The handler self-asserts its MEDIA / MICROPUB_DB / AUTH_DB /
+// TOKEN_SIGNING_KEY bindings on each request and validates IndieAuth-issued
+// access tokens with TOKEN_SIGNING_KEY — the same secret @dwk/indieauth signs
+// them with.
+const micropubByEnv = new WeakMap();
+function micropubFor(env) {
```
davidwkeith, line 264: If `env.SITE_URL` is undefined (e.g. an existing site that hasn't re-run the skill to add the var), `createMicropub` will throw `"baseUrl is not a valid URL"` deep inside the package. That error surfaces in Worker logs but gives the operator no hint that a `SITE_URL` var is missing.

Consider a fast-fail guard:

```js
const origin = env.SITE_URL;
if (!origin) {
  return new Response(
    "SITE_URL is not configured — set it in wrangler.jsonc vars and redeploy",
    { status: 503 }
  );
}
```

(Or just log the clear message and let the package throw — either is better than a stack trace referencing package internals.)

---

tests/__stubs__/dwk-micropub.ts
davidwkeith, line 1: The stub correctly drops `queue` and `scheduled` — `site-entry.js` never calls `micropub.queue()` or `micropub.scheduled()` directly (it delegates to `indieweb-bridge.sync()` instead).

One thing worth confirming: the dispatch tests that check `/micropub` and `/media` routing use this stub's `x-handler: micropub` response header to identify the right handler. The stub is now a plain function (no `.queue`/`.scheduled`). Double-check that the dispatch tests don't call `micropubFor(env).queue(...)` — if any test path tries to, it will throw `TypeError: micropubFor(...).queue is not a function` with no clear message.

---

skills/indieweb/SKILL.md
```diff
@@ -226,10 +226,10 @@ openssl rand -hex 32
 Store it as a wrangler secret on the site Worker (use the `name` value from `wrangler.jsonc`):
 
 ```sh
-npx wrangler secret put INDIEAUTH_SIGNING_KEY --name <CF_PROJECT_NAME>
+npx wrangler secret put TOKEN_SIGNING_KEY --name <CF_PROJECT_NAME>
 ```
 
-Tell the owner to paste the generated key when prompted. The key never appears in source code or `.site-config` — it lives only in Cloudflare's secret store.
+The binding **must** be named `TOKEN_SIGNING_KEY`: `@dwk/indieauth` signs access tokens with it and `@dwk/micropub` verifies them with the same secret, so both endpoints share this one key. Provision it whenever IndieAuth **or** Micropub is selected (they're a token issuer/consumer pair). Tell the owner to paste the generated key when prompted. The key never appears in source code or `.site-config` — it lives only in Cloudflare's secret store.
```
davidwkeith, line 232: **Missing migration path for existing deployments.** Sites that already ran `/anglesite:indieweb` have `INDIEAUTH_SIGNING_KEY` stored as a Cloudflare secret. After updating to this plugin version, both `@dwk/indieauth` and `@dwk/micropub` will read `TOKEN_SIGNING_KEY` — which won't exist — and return 500 on the first request.

The skill needs a migration note here. Suggested addition after this paragraph:

> **Upgrading from a previous version?** If you already have `INDIEAUTH_SIGNING_KEY` set, retrieve its current value from the Cloudflare dashboard (Workers > your project > Settings > Variables & Secrets) and re-provision it under the new name:
> ```sh
> npx wrangler secret put TOKEN_SIGNING_KEY --name <CF_PROJECT_NAME>
> ```
> The old `INDIEAUTH_SIGNING_KEY` binding can be removed once `TOKEN_SIGNING_KEY` is confirmed working.

Without this, existing IndieAuth/Micropub sites go dark silently on the next deploy.

---

davidwkeith (commented):
The core logic is sound — the import rename, per-env memoization of the Micropub handler, and the `bridge_sync` side-table approach are all correct. The test rewrite tracks the schema change well. Four issues to address before merging:

**Must fix:**
1. **Migration path missing** (`skills/indieweb/SKILL.md`) — existing sites with `INDIEAUTH_SIGNING_KEY` will 500 on the next deploy because `TOKEN_SIGNING_KEY` won't exist. Needs an upgrade note with `wrangler secret put TOKEN_SIGNING_KEY`.

**Should fix:**
2. **Infinite cron retry for bad-slug rows** (`indieweb-bridge.js`) — posts whose URL doesn't match `/notes/<slug>/` are skipped with `continue` without any `bridge_sync` record, so the unsynced query picks them up on every run. A sentinel value or a permanent-failure record would stop the log spam.
3. **Error message drops `INDIEAUTH_SIGNING_KEY`** (`scripts/pre-deploy-check.sh`) — the grep still catches the legacy binding name but the human-readable error message doesn't list it, making the failure confusing for owners of legacy sites.

**Nice to fix:**
4. **No `SITE_URL` guard in `micropubFor`** (`site-entry.js`) — if the var is missing, the package throws an opaque error. A fast-fail with a clear message would save debugging time.

---

Issue #381, opened 2026-06-22
apply_edit rejects op=\"apply-instruction\" — the app's Foundation Models chat path sends it expecting plugin-side interpretation


davidwkeith:
## Summary

The plugin's `apply_edit` MCP tool accepts a **closed** op enum (`server/apply-edit-schema.mjs`):

```js
export const editOps = ["replace-text", "replace-attr", "replace-image-src", "edit-style"];
```

`Anglesite/Anglesite-app` has a production path that calls `apply_edit` with **`op: "apply-instruction"`**, which fails our Zod validation. This was confirmed by tracing the app's Swift end-to-end (not speculation).

## Evidence — the op reaches MCP unresolved

Full production call path in `Anglesite/Anglesite-app` (`main`):

1. `Sources/AnglesiteCore/ApplyEditTool.swift` (on-device Foundation Models chat tool, app #251) maps the model's operation straight to the wire string, with no resolution:
   ```swift
   case .applyInstruction: return EditMessage.Op.applyInstruction   // "apply-instruction"
   ```
2. `Sources/AnglesiteCore/IntentEditBridge.swift` forwards it verbatim — no branch on the op:
   ```swift
   let message = EditMessage(… op: op, value: value, dryRun: dryRun)
   return await router.apply(message)
   ```
3. `Sources/AnglesiteCore/MCPApplyEditRouter.swift` ships it to our MCP server unchanged:
   ```swift
   let args = message.jsonValue          // includes "op": "apply-instruction"
   let result = try await toolCaller("apply_edit", args)
   ```

The app's own doc-comment on the constant (`EditMessage.swift`) states the contract:
> `"apply-instruction"` — natural-language edit **forwarded to the plugin for interpretation**. Used by Foundation Models' chat `ApplyEditTool` (#251). Siri AI's `EditContentIntent` (B.5 / #149) now emits concrete ops instead.

So only the **Siri `EditContentIntent`** path pre-resolves to concrete ops. The **Foundation Models chat** path deliberately delegates interpretation to us and currently gets a validation refusal.

## Op-set misalignment (reconcile both directions)

| op | Plugin `apply_edit` | App `EditMessage.Op` |
|---|---|---|
| `replace-text` | ✅ | ✅ |
| `replace-attr` | ✅ | ✅ |
| `replace-image-src` | ✅ | ✅ |
| `edit-style` | ✅ | ❌ (no app sender) |
| `apply-instruction` | ❌ | ✅ (chat path) |

## Options

[...]

Issue #380, opened 2026-06-22
Refresh the Anglesite-app integration contract docs (stale server/ tree, undocumented MCP tools + HTTP transport, missing ADR)


davidwkeith:
## Problem

`Anglesite/Anglesite-app` treats the plugin as the source of truth for the MCP message schema, tools, and hooks, and its developers read our docs as the contract. Several of those docs have fallen behind the code, so the app side is working from a stale picture.

### 1. Root `CLAUDE.md` `server/` tree is out of date
The "Plugin structure" listing for `server/` shows only `annotations.mjs`, `selector.mjs`, `messages.mjs`, `apply-edit-schema.mjs`, `patcher.mjs`, `index.mjs`. The actual directory also contains: `index-tools.mjs`, `http-server.mjs`, `apply-edit-dispatcher.mjs`, `style-edit.mjs`, `undo-edit.mjs`, `edit-history.mjs`, `create-content.mjs`, `list-content.mjs`, `content-frontmatter.mjs`, `optimize-images.mjs`.

It also doesn't document:
- the **8 MCP tools** now registered (`add_annotation`, `list_annotations`, `resolve_annotation`, `apply_edit`, `undo_edit`, `list_content`, `create_page`, `create_post`), or
- the **HTTP/Streamable transport** (`ANGLESITE_MCP_TRANSPORT`/`_HOST`/`_PORT`, `/mcp`, `Mcp-Session-Id`, the `"Anglesite MCP listening on …"` readiness line) the app uses for container runtimes.

### 2. `docs/dev/mac-app-design.md` is a pre-edit-pipeline draft
Marked **"Status: Draft / exploration."** §5.4 still says the server "handles three messages" (annotations only) and predates the entire apply_edit pipeline, content tools, and HTTP transport that have since shipped. §14 says the design "should be ratified as a new ADR (`0020-native-mac-app.md`)" — but **0020 is already taken** (`0020-active-indieweb.md`) and no mac-app ADR was ever created.

### 3. `server/apply-edit-schema.mjs:11` op-enum comment drift
The header comment still says `op` is "the closed enum {replace-text, replace-attr, replace-image-src}", but `editOps` (line 41) now includes **`edit-style`**.

## Proposed fix

- [ ] Update the `server/` tree + add an "MCP tools" and "MCP transport (stdio + HTTP)" subsection to root `CLAUDE.md`.
[...]

Issue #379, opened 2026-06-22
MCP server reports stale version (0.16.4) — sync with plugin version on release


davidwkeith:
## Problem

`server/index-tools.mjs` hard-codes the MCP server identity:

```js
const server = new McpServer({
  name: "anglesite-annotations",
  version: "0.16.4",
});
```

The plugin is at **1.2.0** (`package.json`, `.claude-plugin/plugin.json`, `template/package.json`). The `Anglesite-app` MCP client reads server name+version during the MCP `initialize` handshake, so the host sees `0.16.4` while the bundled plugin it copied is `1.2.0`. The app also stamps the bundled plugin with its origin commit (`Resources/plugin/.bundled-from-commit`), so this mismatch is observable.

The documented version-sync process (`CLAUDE.md` → Version management; `bin/release.ts`) keeps **three** files in lockstep but does **not** touch the MCP server version, so it has drifted.

## Proposed fix

Either:
- Have `bin/release.ts` also rewrite the `version:` in `server/index-tools.mjs`, and add it to the "versions must stay in sync" list in `CLAUDE.md`; **or**
- Read the version from `.claude-plugin/plugin.json` / `package.json` at server startup instead of hard-coding it (single source of truth).

The second is more robust. Also consider whether `name: "anglesite-annotations"` should be renamed now that the server is much more than annotations (apply_edit pipeline, content tools) — though that's a wire-visible identifier the app may key on, so coordinate with the app team before changing it.

## Context

Found during a plugin↔app support review.

Issue #378, opened 2026-06-22
CI: test on Node 24.x (the Anglesite-app vendored runtime), not just Node 22


davidwkeith:
## Problem

The plugin declares `"engines": { "node": ">=22" }` and CI (`.github/workflows/test.yml`, `release.yml`) pins **Node 22** for both lint/test and release.

But `Anglesite/Anglesite-app` vendors and ships a specific Node runtime — `scripts/node-version.txt` = **`24.15.0`** (arm64) — and that is the interpreter that actually executes the plugin's `server/*.mjs` MCP server and `template/scripts/*.ts` in production (both the embedded-Node path and the OCI container image pin to that file).

So the runtime we ship to end users via the app is **never exercised in CI**. `>=22` is satisfied, but a Node-24-only regression (e.g. a deprecation, a changed default, an ESM/`node:` API difference) would pass CI and break the app.

## Proposed fix

Add Node 24.x to the test matrix in `.github/workflows/test.yml` (keep 22 as the floor):

```yaml
strategy:
  matrix:
    node-version: [22, 24]
```

Optionally bump the release workflow to build/verify under the same Node the app pins. Consider documenting the coupling in `CLAUDE.md` (Version management) so the matrix tracks `Anglesite-app/scripts/node-version.txt`.

## Context

Found during a plugin↔app support review. The app's MCP client spawns `server/index.mjs` under the vendored Node; the container runtime (#59/#61/#69 in the app repo) bakes the same Node + plugin server into a `linux/arm64` image.

PR #382, opened 2026-06-23
chore: sync MCP server version, CI Node 24, and Anglesite-app contract docs

davidwkeith:
Quick-win fixes from the plugin↔`Anglesite-app` integration review. Closes the three low-risk gaps; the one functional gap (`apply_edit` rejecting `op="apply-instruction"`) is tracked separately in #381 and **not** addressed here.

### #378 — CI tests on Node 24
`.github/workflows/test.yml`: the `test` job now runs a matrix of Node **22 and 24**. `22` is the `engines` floor; **24** is the runtime `Anglesite-app` vendors and ships (`scripts/node-version.txt`) to actually execute `server/*.mjs` in production — previously never exercised in CI.

### #379 — MCP server version no longer drifts
`server/index-tools.mjs` read a hard-coded `version: "0.16.4"` while the plugin is at `1.2.0`. It now reads the version from `.claude-plugin/plugin.json` at startup, so the MCP `initialize` handshake always tracks the plugin version. No new file to bump in `bin/release.ts`; documented in `CLAUDE.md` → Version management.

### #380 — Refresh the Anglesite-app contract docs
- **Root `CLAUDE.md`**: rewrote the stale `server/` tree (was missing 10 modules) and added an **"MCP server & the Anglesite-app host"** section documenting the 8 MCP tools and the stdio/HTTP transport contract.
- **`server/apply-edit-schema.mjs`**: op-enum header comment now includes `edit-style`.
- **`docs/dev/mac-app-design.md`**: status flipped from "Draft / exploration" to "Shipped", §5.4 corrected from "three messages" to the as-built 8-tool/transport reality, and the dangling `0020` ADR reference fixed.
- **ADR-0023** (`docs/decisions/0023-native-mac-app.md`): ratifies the native-host decision (the design doc's suggested `0020` was already taken by `0020-active-indieweb.md`); README index updated (also backfilled the missing `0022`).

## Testing

`npm test` — all green **except** 2 pre-existing failures in `test/apply-edit-dispatcher.test.js` that are **unrelated to this PR**: they simulate a locked filesystem via `chmod 0o555` on a directory, which root (uid 0, as this container runs) bypasses. They pass on GitHub Actions' non-root `ubuntu` runner. Verified pre-existing by stashing this branch's changes and re-running.

## Out of scope (tracked separately)
- #381 — `apply_edit` rejecting `op="apply-instruction"` (functional; needs a design decision)
- Op-set reverse gap: `edit-style` has no app sender (informational)

https://claude.ai/code/session_01P1guSM3tVAUT7bcWwQNheJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01P1guSM3tVAUT7bcWwQNheJ)_

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,12 +9,17 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # 22 = the engines floor; 24 = the runtime Anglesite-app vendors and ships
+        # (scripts/node-version.txt), which actually executes server/*.mjs in production.
+        node-version: [22, 24]
     steps:
       - uses: actions/checkout@v4
 
       - uses: actions/setup-node@v4
         with:
-          node-version: 22
+          node-version: ${{ matrix.node-version }}
 
       - run: npm ci
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,6 +10,9 @@ jobs:
   test:
     runs-on: ubuntu-latest
     strategy:
+      # Run both legs to completion so a regression on one Node version doesn't
+      # mask the other — the whole point of the matrix.
+      fail-fast: false
       matrix:
         # 22 = the engines floor; 24 = the runtime Anglesite-app vendors and ships
         # (scripts/node-version.txt), which actually executes server/*.mjs in production.
--- a/server/index-tools.mjs
+++ b/server/index-tools.mjs
@@ -1,6 +1,26 @@
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
 import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import { StdioServerTransport } from "@modelcontextprotocol/sdk/server/stdio.js";
 import { z } from "zod";
+
+/**
+ * The plugin version is the single source of truth (`bin/release.ts` keeps the
+ * three manifests in lockstep). Read it from the manifest at startup so the MCP
+ * `initialize` handshake never drifts from the bundled plugin the host copied.
+ */
+function pluginVersion() {
+  try {
+    const manifestUrl = new URL(
+      "../.claude-plugin/plugin.json",
+      import.meta.url,
+    );
+    const manifest = JSON.parse(readFileSync(fileURLToPath(manifestUrl), "utf8"));
+    return manifest.version ?? "0.0.0";
+  } catch {
+    return "0.0.0";
+  }
+}
 import {
   addAnnotation,
   listAnnotations,
@@ -20,7 +40,7 @@ import { createPage, createPost } from "./create-content.mjs";
 export function buildServer(projectRoot) {
   const server = new McpServer({
     name: "anglesite-annotations",
-    version: "0.16.4",
+    version: pluginVersion(),
   });
 
   server.tool(
--- a/server/index-tools.mjs
+++ b/server/index-tools.mjs
@@ -3,6 +3,17 @@ import { fileURLToPath } from "node:url";
 import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import { StdioServerTransport } from "@modelcontextprotocol/sdk/server/stdio.js";
 import { z } from "zod";
+import {
+  addAnnotation,
+  listAnnotations,
+  resolveAnnotation,
+} from "./annotations.mjs";
+import { applyEditInputShape } from "./apply-edit-schema.mjs";
+import { applyEdit } from "./apply-edit-dispatcher.mjs";
+import { recordEdit } from "./edit-history.mjs";
+import { undoEdit } from "./undo-edit.mjs";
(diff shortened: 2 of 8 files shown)


server/index-tools.mjs
```diff
@@ -1,6 +1,26 @@
...
+    const manifestUrl = new URL(
+      "../.claude-plugin/plugin.json",
+      import.meta.url,
+    );
+    const manifest = JSON.parse(readFileSync(fileURLToPath(manifestUrl), "utf8"));
+    return manifest.version ?? "0.0.0";
+  } catch {
+    return "0.0.0";
+  }
+}
```
davidwkeith, line 37: The `pluginVersion()` function is defined between two import blocks — the SDK imports (`McpServer`, `z`, etc.) end at line 5, then the function runs lines 12–23, and the local module imports pick up at line 24. ESM static imports are hoisted so this is technically fine, but it violates the convention that all imports come before any declarations, and most linters (`import/first`) would flag it.

Move the function definition below all the imports (i.e. after `import { createPage, createPost }`) so the file reads: all imports → `pluginVersion` → `buildServer`.

---

server/index-tools.mjs
```diff
@@ -1,6 +1,26 @@
...
+function pluginVersion() {
+  try {
+    const manifestUrl = new URL(
+      "../.claude-plugin/plugin.json",
+      import.meta.url,
+    );
+    const manifest = JSON.parse(readFileSync(fileURLToPath(manifestUrl), "utf8"));
+    return manifest.version ?? "0.0.0";
+  } catch {
+    return "0.0.0";
```
davidwkeith, line 35: Silent fallback. If `plugin.json` is missing or malformed (e.g. wrong install path, bundled copy wasn't updated), the server silently advertises itself as `"0.0.0"` during the MCP `initialize` handshake — which the app-host may use for compatibility checks. A `console.warn` here would surface the problem without breaking startup:

```js
} catch (err) {
  console.warn("[anglesite] could not read plugin version:", err.message);
  return "0.0.0";
}
```

---

server/index-tools.mjs
```diff
@@ -20,7 +40,7 @@ import { createPage, createPost } from "./create-content.mjs";
 export function buildServer(projectRoot) {
   const server = new McpServer({
     name: "anglesite-annotations",
```
davidwkeith: The server `name` is still `"anglesite-annotations"` but this PR's own new `CLAUDE.md` section calls out 8 tools spanning annotations, edits, content creation, and undo. The app host receives this name in the MCP `initialize` response — if it ever matches on it for routing or display, the mismatch will be confusing. Worth updating to something like `"anglesite"` to match the plugin name.

---

.github/workflows/test.yml
```diff
@@ -9,12 +9,17 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # 22 = the engines floor; 24 = the runtime Anglesite-app vendors and ships
+        # (scripts/node-version.txt), which actually executes server/*.mjs in production.
+        node-version: [22, 24]
```
davidwkeith, line 19: Without `fail-fast: false`, a failure on Node 22 will cancel the Node 24 run (and vice versa). The whole point of the matrix is to see which version behaves differently, so you probably want both legs to always run to completion:

```yaml
strategy:
  fail-fast: false
  matrix:
    node-version: [22, 24]
```

---

davidwkeith (commented):
The three doc changes (ADR-0023, `mac-app-design.md` updates, CLAUDE.md server section) and the Node 24 matrix are all good. Four things to fix before merge:

1. **Import ordering in `server/index-tools.mjs`** — `pluginVersion()` is inserted between SDK imports and local module imports. Move it below all imports.
2. **Silent `"0.0.0"` fallback** — the `catch` swallows the error; add a `console.warn` so a bad bundled path doesn't go unnoticed.
3. **Stale server name** — `"anglesite-annotations"` should be updated now that the tool set covers edits and content too.
4. **Missing `fail-fast: false`** in the CI matrix — without it, a Node 22 failure cancels the Node 24 run, defeating the purpose of the matrix.

See inline comments for specifics.

---

davidwkeith:
All four addressed in `afd08d9`:

1. ✅ `pluginVersion()` moved below all imports.
2. ✅ `catch (err)` now `console.warn`s (stderr — doesn't touch the stdio JSON-RPC channel; the `initialize` test still parses clean).
3. ✅ Server `name` → `"anglesite"`, and the `serverInfo.name` assertion in `tests/mcp-server.test.ts` updated to match. One scoping note: I left the client-side `.mcp.json` registration key and the toolbar `id` (`"anglesite-annotations"`) **unchanged** — those are a matched pair the scaffolded toolbar wires against, separate from the server's self-reported `initialize` name, and outside this review's scope. Worth a quick check that the app-host doesn't match on the old `serverInfo.name` for routing/display before this ships.
4. ✅ `fail-fast: false` added to the matrix.

CI re-running on the new commit.

---

PR #383, opened 2026-06-23
Add Open Agent Skills (skills.sh) export alongside the plugin

davidwkeith:
## Summary

Generates a spec-compliant [Open Agent Skills](https://agentskills.io) ([skills.sh](https://www.skills.sh)) export of all 56 skills under `agent-skills/`, **coexisting** with the existing Claude Code plugin. Skills become installable with:

```sh
npx skills add Anglesite/anglesite/agent-skills/<skill>
```

`skills/` remains the source of truth; `agent-skills/` is generated and committed so the skills.sh CLI can resolve skills by path.

## Why a build step

A single file can't satisfy both formats — `${CLAUDE_PLUGIN_ROOT}` is **required** by Claude Code but **fatal** to the spec (no variable substitution). So `bin/build-agent-skills.ts` transforms the plugin sources:

| Plugin feature | Spec status | Transform |
|---|---|---|
| `${CLAUDE_PLUGIN_ROOT}/...` (317 refs) | unsupported | rewritten to relative `references/<path>`; file bundled |
| `disable-model-invocation`, `user-invocable` | not in spec | → `metadata.invocation` |
| `argument-hint` | not in spec | → `metadata.argument-hint` |
| cross-skill `SKILL.md` links (15 skills) | no mechanism | → plain mention (``the `x` skill``) |

Frontmatter is normalized to the spec (`name`, `description`, `license`, `compatibility`, `allowed-tools`, `metadata`); names/descriptions are validated against the spec rules.

## What's included

- **`bin/build-agent-skills.ts`** + `npm run build:agent-skills`
- **`agent-skills/`** — 56 generated skills with bundled `references/` and a generated index `README.md`
- **`tests/build-agent-skills.test.ts`** — 68 tests covering the transforms and asserting the committed export matches a fresh build
- **CI sync check** in `.github/workflows/test.yml` (fails if `agent-skills/` is stale)
- **`docs/dev/agent-skills.md`** — full contract + limitations
- Fixed two pre-existing broken ADR links in the source (`design-import`, `print`)

## Known limitations (documented, surfaced as build warnings)

- **Self-containment tradeoff:** skills referencing `docs/smb/<BUSINESS_TYPE>.md` bundle the whole 66-file business-guide catalog, so several skills duplicate it (~11 MB total). A leaner reference-without-bundling mode is possible if preferred.
- **Nested refs:** a few bundled docs (mainly the `import` skill) still contain their own `${CLAUDE_PLUGIN_ROOT}` references (not transitively rewritten).
- **Body length:** `convert`, `deploy`, `import` exceed the spec's recommended 500-line budget — functional, candidates for splitting.
- **Genuine source bug flagged, not fixed:** `design-import` invokes `scripts/design-import/compare-screenshots.mjs`, which doesn't exist (the real module is `comparison.mjs`, a programmatic export, not a CLI). Left for a maintainer decision.
- **Registry listing:** `npx skills add <path>` works straight from git; getting indexed by `npx skills find` is a separate skills.sh submission.

## Test plan

- `npm run build:agent-skills` → 7 informational warnings, 0 validity failures
- `npx vitest run tests/build-agent-skills.test.ts` → 68 passed
- Full suite: only 2 pre-existing failures in `test/apply-edit-dispatcher.test.js` (root bypasses the non-writable-dir check; fail on `main` too)

---
_Generated by [Claude Code](https://claude.ai/code/session_01BMWSoPonFrhPnsfEfPyDg1)_


bin/build-agent-skills.ts
```diff
@@ -0,0 +1,454 @@
...
+    if (bundled.has(relPath)) return;
+    const src = join(pluginRoot, relPath);
+    const dest = join(skillDir, "references", relPath);
+    mkdirSync(dirname(dest), { recursive: true });
+    const st = statSync(src);
+    cpSync(src, dest, { recursive: st.isDirectory() });
+    bundled.add(relPath);
+    // Flag nested plugin-root references inside copied text files (spec
+    // discourages deep reference chains; these won't resolve standalone).
+    if (st.isFile() && PLUGIN_VAR_RE.test(readFileSync(src, "utf-8"))) {
```
davidwkeith: **Bug: stale `lastIndex` on the global regex.** `PLUGIN_VAR_RE` is a module-level `/.../g` regex. `test()` advances `lastIndex` to the position after the first match and leaves it there on a truthy return. When `copyInto` is called for a _second_ reference file in the same skill (e.g. the `import` skill bundles several docs), the next `PLUGIN_VAR_RE.test()` starts from the previous `lastIndex` — and if `${CLAUDE_PLUGIN_ROOT}` appears earlier in the new string, the nested-ref warning is silently skipped.

Notably, `rewriteBody` already guards its while-loop with `REF_RE.lastIndex = 0` — the same pattern is needed here:

```ts
PLUGIN_VAR_RE.lastIndex = 0;
if (st.isFile() && PLUGIN_VAR_RE.test(readFileSync(src, "utf-8"))) {
```

Or, simpler, use a non-global regex for a single-occurrence check:
```ts
if (st.isFile() && /\$\{CLAUDE_PLUGIN_ROOT\}/.test(readFileSync(src, "utf-8"))) {
```

---

tests/build-agent-skills.test.ts
davidwkeith, line 1: `emitSkill` has no unit tests. The 68 tests cover `parseSkill`, `rewriteBody`, `buildFrontmatter`, `validateSkill`, and `renderIndex` well, and the sync guard catches stale committed output — but the actual file-writing, reference bundling, and nested-ref detection logic in `emitSkill` is exercised only indirectly through the full build.

Worth adding at minimum:
- A test that confirms `emitSkill` emits the `NESTED REF:` warning when a bundled file contains `${CLAUDE_PLUGIN_ROOT}` (this would have caught the `lastIndex` bug above).
- A test for the `DYNAMIC REF:` and `MISSING REFERENCE:` warning paths.

---

.github/workflows/test.yml
davidwkeith, line 1: If PR #382 (Node 22 + 24 matrix) lands before this PR, the `Verify Open Agent Skills export is up to date` step runs on **both** Node versions — regenerating 300+ files twice per push. The check is deterministic and has nothing to do with the Node version, so this is pure waste.

Fix: either move it to a separate job outside the matrix, or guard it:
```yaml
- name: Verify Open Agent Skills export is up to date
  if: matrix.node-version == 22
  run: |
    ...
```

---

agent-skills/README.md
davidwkeith, line 1: Two things missing from this generated index:

1. **`linguist-generated` attribute** — 300+ files in `agent-skills/` will inflate GitHub's language stats and make every PR touching `skills/` show a wall of generated diffs. Add to `.gitattributes`:
   ```
   agent-skills/ linguist-generated=true
   ```

2. **Registry caveat** — The PR description notes that `npx skills find` (the skills.sh registry) isn't set up yet; only direct-path `npx skills add` works. The README should say so to avoid confusion:
   > Note: Anglesite is not yet listed in the skills.sh public registry. Install directly by path as shown above; `npx skills find anglesite` will not return results until a registry submission is made.

(The `.gitattributes` comment belongs in a standalone thread but there's no existing file to attach it to — adding it here.)

---

davidwkeith (commented):
Overall concept is sound — the coexistence approach (skills/ stays the source of truth, agent-skills/ generated and committed) is clean, and the test coverage for the transform logic is thorough. Four issues to resolve before merging:

1. **Bug: `PLUGIN_VAR_RE.lastIndex` not reset between `test()` calls in `emitSkill`** — when a skill has multiple bundled reference files (e.g. `import`, `convert`), the second file's nested-ref check starts from the wrong position and can silently miss `${CLAUDE_PLUGIN_ROOT}` occurrences. See inline comment on line 320. `rewriteBody` already guards `REF_RE.lastIndex = 0` for exactly this reason — same fix needed here.

2. **`emitSkill` not unit-tested** — the 68 tests cover the transform logic well but none exercise `emitSkill` directly. The nested-ref warning path (the bug above) and the `DYNAMIC REF:` / `MISSING REFERENCE:` paths are only exercised by the sync guard, which verifies committed output, not the warning logic. See inline comment on the test file.

3. **CI sync check will run twice** if PR #382's Node 22+24 matrix lands first. The check is deterministic and Node-version-independent. See inline comment on `test.yml`.

4. **`.gitattributes` missing** — 300+ generated files need `agent-skills/ linguist-generated=true` to avoid inflating language stats and polluting PR diffs. See inline comment on `agent-skills/README.md`.

Also noted (non-blocking, but worth tracking):
- The `design-import` source bug (invokes `compare-screenshots.mjs` which doesn't exist) is flagged in the PR description but has no `# TODO` in the source skill itself. A future maintainer generating a fresh export won't know this warning is expected.
- `convert`, `deploy`, and `import` exceed the 500-line budget. Functional, but candidates for splitting.

---

davidwkeith:
Thanks for the thorough review — all four addressed in ff4a8c9:

1. **Stale `lastIndex` bug** — switched the nested-ref check to a non-global regex. This was a real silent miss: the build now reports **3** `NESTED REF` warnings for `import` instead of 2 (`skills/import/newsletter-migration.md` was being skipped).
2. **`emitSkill` unit tests** — added 5 tests covering the NESTED REF path (multi-file, which reproduces the bug above), plus the `DYNAMIC REF`, `MISSING REFERENCE`, and successful-bundle paths.
3. **CI duplicate work** — moved the export sync check into its own `agent-skills-export` job, so it runs once regardless of any Node matrix in #382.
4. **`.gitattributes`** — added `agent-skills/** linguist-generated=true`, and added the registry caveat (`npx skills find` not yet available) to the generated `README.md`.

Non-blocking notes: added a `<!-- TODO -->` in `design-import`'s source next to the `compare-screenshots.mjs` invocation so the expected `MISSING REFERENCE` warning is explained at the source. Left the 500-line splitting of `convert`/`deploy`/`import` as a tracked follow-up in `docs/dev/agent-skills.md`.

---

PR #384, opened 2026-06-23
fix(apply-edit): accept op=apply-instruction with structured needs-agent refusal

### LINKED ISSUE #381: apply_edit rejects op=\"apply-instruction\" — the app's Foundation Models chat path sends it expecting plugin-side interpretation

davidwkeith:
## Summary

- Adds `apply-instruction` to the `apply_edit` op enum so the Anglesite-app's Foundation Models chat path (#251) no longer gets a Zod validation crash
- Returns a structured `edit-failed` response with `reason: "needs-agent"` instead, so the app can cleanly route the NL instruction to its agent/chat path
- Fixes the stale op-enum comment in `apply-edit-schema.mjs` (from #380)

## Test plan

- [x] New schema test: `apply-instruction` accepted by Zod enum
- [x] New dispatcher test: returns `edit-failed` with `reason: "needs-agent"` and `isError: true`
- [x] New dispatcher test: filesystem untouched (no patcher invocation)
- [x] All existing `editOps` enum tests still pass (including `accepts every enum op`)
- [x] Full test suite passes (2 pre-existing `write-failed` failures due to root user in CI container — unrelated)

Closes #381

https://claude.ai/code/session_01NqrLR2UQuzRNVXphAoQbmF

---
_Generated by [Claude Code](https://claude.ai/code/session_01NqrLR2UQuzRNVXphAoQbmF)_

### DIFF
--- a/server/apply-edit-schema.mjs
+++ b/server/apply-edit-schema.mjs
@@ -61,12 +61,12 @@ export const applyEditInputShape = {
   op: z
     .enum(editOps)
     .describe(
-      "Edit operation: replace-text (innerText), replace-attr (value is {name, value}), replace-image-src (value is {filename, mimeType, dataURL}), edit-style (value is {property, value}; merges a rule into the owning component's scoped <style>), apply-instruction (NL instruction forwarded by the app's Foundation Models chat — returns needs-agent for app-side routing)",
+      "Edit operation: replace-text (innerText), replace-attr (value is {name, value}), replace-image-src (value is {filename, mimeType, dataURL}), edit-style (value is {property, value}; merges a rule into the owning component's scoped <style>), apply-instruction (reserved: sent only by the Anglesite-app Foundation Models chat path; always returns edit-failed/needs-agent — do not use from external callers)",
     ),
   value: z
     .unknown()
     .describe(
-      "Operation payload; varies by op (string for replace-text, {name, value} for replace-attr, {filename, mimeType, dataURL} for replace-image-src, {property, value} for edit-style)",
+      "Operation payload; varies by op (string for replace-text, {name, value} for replace-attr, {filename, mimeType, dataURL} for replace-image-src, {property, value} for edit-style, string for apply-instruction — the NL instruction text)",
     ),
   dry_run: z
     .boolean()
--- a/server/apply-edit-schema.mjs
+++ b/server/apply-edit-schema.mjs
@@ -8,7 +8,7 @@
  *   - `selector: ElementInfo` (drops `element/tagName/textFingerprint/domPath`; matches
  *     selector.mjs's existing typedef).
  *   - `path` (drops `url`; the app already uses `path` on its side).
- *   - `op` is the closed enum {replace-text, replace-attr, replace-image-src, edit-style}.
+ *   - `op` is the closed enum {replace-text, replace-attr, replace-image-src, edit-style, apply-instruction}.
  *   - No `site` field — the MCP server already knows its `projectRoot`.
  *   - `type` is accepted-and-ignored — the app uses it as a WKWebView-side boundary tag.
  */
@@ -36,9 +36,10 @@ export const elementInfoSchema = z.object({
   textContent: z.string().optional(),
 });
 
-/** Edit operations the patcher knows how to apply. Phase 5's patcher (#295) implements these;
- *  new ops require an explicit enum addition + a resolver update. */
-export const editOps = ["replace-text", "replace-attr", "replace-image-src", "edit-style"];
+/** Edit operations the server accepts. The patcher resolves the first four; `apply-instruction`
+ *  is an NL-forwarding op the app's Foundation Models chat path sends — the server returns a
+ *  structured `needs-agent` refusal so the app can route it to its agent. */
+export const editOps = ["replace-text", "replace-attr", "replace-image-src", "edit-style", "apply-instruction"];
 
 /** The MCP tool's input shape, as passed to `server.tool(name, description, shape, handler)`. */
 export const applyEditInputShape = {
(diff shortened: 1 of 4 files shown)


server/apply-edit-schema.mjs
```diff
@@ -60,7 +61,7 @@ export const applyEditInputShape = {
   op: z
     .enum(editOps)
     .describe(
-      "Edit operation: replace-text (innerText), replace-attr (op needs value to be {name, value}), replace-image-src (value is {filename, mimeType, dataURL}), edit-style (value is {property, value}; merges a rule into the owning component's scoped <style>)",
+      "Edit operation: replace-text (innerText), replace-attr (value is {name, value}), replace-image-src (value is {filename, mimeType, dataURL}), edit-style (value is {property, value}; merges a rule into the owning component's scoped <style>), apply-instruction (NL instruction forwarded by the app's Foundation Models chat — returns needs-agent for app-side routing)",
```
davidwkeith: The `op` describe string is part of the MCP tool schema that gets surfaced to every LLM client. Advertising `apply-instruction` there — with a description of how it routes — could prompt those clients to use it expecting NL editing, only to get a `needs-agent` refusal in a loop.

Consider framing it as explicitly app-internal so third-party callers know not to touch it:

```diff
- apply-instruction (NL instruction forwarded by the app's Foundation Models chat — returns needs-agent for app-side routing)
+ apply-instruction (reserved: sent only by the Anglesite-app Foundation Models chat path; always returns edit-failed/needs-agent — do not use from external callers)
```

Or, if keeping the current wording, at least lead with "do not call from external clients" so the refusal is unsurprising.

---

test/apply-edit-dispatcher.test.js
```diff
@@ -140,6 +140,31 @@ describe("applyEdit — write-failed surface", () => {
...
+  it("returns edit-failed with reason needs-agent instead of crashing", async () => {
+    const response = await applyEdit(root, makeEdit({
+      op: "apply-instruction",
+      value: "Make the heading more welcoming",
+    }));
+    expect(response.isError).toBe(true);
+    const body = parseContent(response);
+    expect(body.type).toBe("anglesite:edit-failed");
+    expect(body.id).toBe("e-1");
+    expect(body.reason).toBe("needs-agent");
```
davidwkeith, line 153: The `failed()` call in the dispatcher passes a `detail` string (`"apply-instruction requires LLM interpretation; route to the agent"`), but the test doesn't assert it. If a future refactor accidentally drops `detail`, nothing would catch it.

Suggest adding one more assertion here:

```js
expect(body.reason).toBe("needs-agent");
expect(body.detail).toBeTruthy(); // or pin to the exact string if you want it locked
```

---

davidwkeith (commented):
The approach is sound — catching `apply-instruction` in Zod rather than crashing, and returning a structured `needs-agent` refusal the app can act on cleanly. Three things to look at before merge:

**1. MCP `describe()` text leaks app-internal routing to all LLM clients (see inline comment on `apply-edit-schema.mjs`)**

The `.describe()` string becomes part of the public MCP tool schema that every LLM client sees. Advertising `apply-instruction` there without a "do not call from external clients" signal could prompt third-party callers to use it expecting NL editing, then get a `needs-agent` refusal they can't act on. Needs an explicit "reserved / do not use externally" note in the description.

**2. Test doesn't assert the `detail` field (see inline comment on `test/apply-edit-dispatcher.test.js`)**

The `failed()` call passes a detail string, but the test only checks `type`, `id`, and `reason`. A future refactor that accidentally drops `detail` would go unnoticed.

**3. Minor: `value` describe text doesn't cover `apply-instruction`**

`applyEditInputShape.value` describes the payload shape for each op but skips `apply-instruction` (which is a plain NL string). Every other op is documented there — worth adding one more entry for consistency.

---

PR #389, opened 2026-06-26
feat(#377): typed content scaffolding in create-content.mjs

davidwkeith:
Closes #377.

Mirrors the app's `ContentScaffold.renderEntry` / `NativeContentOperations.createTyped` (V-1.2, #344) into the Node MCP sidecar so the MCP create backend reaches typed-content parity with the native path. Until now the MCP path could only scaffold page/post; external MCP consumers (chat assistant, Claude Code CLI) couldn't create typed entries.

## Changes
- **`server/content-types.mjs`** (new) — built-in content-type registry mirrored from the Swift `ContentTypeRegistry` catalog (scaffolding-relevant subset: `id`, `displayName`, `storage`, ordered `fields`). Swift remains the source of truth; mf2/schema.org projections are intentionally not mirrored (scaffolding never reads them).
- **`server/create-content.mjs`** — `createTyped` + `renderEntry`, byte-faithful to the Swift scaffolder: same ISO-8601 date format (`Date#toISOString` ≡ `.withFractionalSeconds`), field-kind defaults, `title`/`name` special-case, markdown-as-body, and the page-stored-type guard (`businessProfile` → not supported yet, #345).
- **`server/index-tools.mjs`** — register the `create_content` MCP tool (`z.enum` over the registry ids).

## Tests
9 new `createTyped` cases in `test/create-content.test.js`:
- Byte-exact frontmatter for note / article / event / like / review (covers markdown-as-body, `name` vs `title` fields, frontmatter-only types, number defaults, YAML escaping, shared `now` across datetime fields).
- Slug fallback to descriptor id on empty title; unknown-type, page-stored, overwrite, and no-git paths.
- Updated the `mcp-server` tool-list assertion to include `create_content`.

Full plugin suite green except one **pre-existing, unrelated** failure (`optimize-images-packaging.test.js` — a sharp/tsx environment issue that fails identically on clean `main`).

## Two-repo follow-on
Per the two-repo protocol, this is the plugin half. The app-side step (bump the bundled-plugin pointer to consume `create_content`) lands separately after this ships in a tagged plugin release.

### DIFF
--- a/server/content-types.mjs
+++ b/server/content-types.mjs
@@ -1,0 +1,177 @@
+/**
+ * Built-in typed content objects (V-1 / epic #334), mirrored from the app's
+ * `AnglesiteCore/ContentTypeRegistry.swift` built-in catalog.
+ *
+ * **Source of truth is Swift.** This is the scaffolding-relevant projection only — each
+ * descriptor's `id`, `displayName`, `storage`, and ordered `fields` (name + kind). The
+ * microformats2 / schema.org projections that the Swift descriptors also carry are NOT
+ * mirrored here: scaffolding (`renderEntry`) never reads them, and the template/JSON-LD
+ * layers that do live app-side. Keep this list byte-faithful to the Swift catalog so the
+ * native and MCP create paths produce identical files (no git churn when the backend swaps).
+ *
+ * Field `kind` values match `ContentTypeField.Kind`: string, text, markdown, bool, date,
+ * datetime, url, image, number, stringArray, imageArray.
+ *
+ * @module
+ */
+
+/**
+ * @typedef {{ name: string, kind: string }} ContentTypeField
+ * @typedef {{ id: string, displayName: string, collection: string | null, fields: ContentTypeField[] }} ContentTypeDescriptor
+ */
+
+/** @type {ContentTypeDescriptor[]} */
+const PERSONAL_TYPES = [
+  {
+    id: "note",
+    displayName: "Note",
+    collection: "notes",
+    fields: [
+      { name: "body", kind: "markdown" },
+      { name: "publishDate", kind: "datetime" },
+      { name: "tags", kind: "stringArray" },
+    ],
+  },
+  {
+    id: "article",
+    displayName: "Article",
+    collection: "articles",
+    fields: [
+      { name: "title", kind: "string" },
+      { name: "summary", kind: "text" },
+      { name: "body", kind: "markdown" },
+      { name: "publishDate", kind: "datetime" },
+      { name: "updated", kind: "datetime" },
+      { name: "tags", kind: "stringArray" },
+    ],
+  },
+  {
+    id: "photo",
+    displayName: "Photo",
+    collection: "photos",
+    fields: [
+      { name: "image", kind: "image" },
+      { name: "caption", kind: "text" },
+      { name: "publishDate", kind: "datetime" },
+      { name: "tags", kind: "stringArray" },
+    ],
+  },
+  {
+    id: "album",
+    displayName: "Album",
+    collection: "albums",
+    fields: [
+      { name: "title", kind: "string" },
+      { name: "images", kind: "imageArray" },
+      { name: "body", kind: "markdown" },
+      { name: "publishDate", kind: "datetime" },
+      { name: "tags", kind: "stringArray" },
+    ],
+  },
+  {
+    id: "bookmark",
+    displayName: "Bookmark",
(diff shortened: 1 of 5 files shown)


davidwkeith (commented):
Good overall structure — the mirroring strategy (Swift is source of truth, JS is the scaffolding-relevant projection) is sound, the tests are thorough, and the byte-fidelity goal is well-documented. A few things to fix before merging.

---

## Bug: `businessProfile` in `z.enum(contentTypeIds)` — schema accepts what the function rejects

**`server/index-tools.mjs`**, `type: z.enum(contentTypeIds)`:

`contentTypeIds` includes `businessProfile`. A caller that passes `type: "businessProfile"` passes schema validation cleanly, then hits `createTyped`'s guard and gets back `{ isError: true, content: [{ text: "Page-stored type businessProfile is not supported by createTyped yet" }] }`. That's a confusing cliff: the tool schema says it's a valid type, but the tool call fails.

Fix: export a second list from `content-types.mjs` and use it in the tool registration:

```js
// content-types.mjs
export const creatableContentTypeIds = contentTypes
  .filter((t) => t.collection !== null)
  .map((t) => t.id);
```

```js
// index-tools.mjs
import { creatableContentTypeIds } from "./content-types.mjs";
// …
type: z.enum(creatableContentTypeIds).describe("…")
```

The `createTyped` guard can stay as a defence-in-depth check, but the MCP schema should no longer advertise an input that will always error.

---

## Bug: `renderEntry` silently drops unknown field kinds

**`server/create-content.mjs`**, `renderEntry` switch:

The `switch` has no `default` case. If the Swift catalog gains a new `kind` (or a typo slips into the registry), that field is silently omitted from the generated frontmatter — no error, no warning, just a missing key. Keystatic will then either error on parse or silently use its field default, which is hard to debug.

```js
default:
  throw new Error(`renderEntry: unhandled field kind "${field.kind}" on field "${field.name}"`);
```

Since `renderEntry` is pure and only called from within this module, throwing is safe — `createTyped` already catches and re-surfaces errors. An alternative is a `console.warn` if you want the scaffold to proceed rather than fail, but at minimum make the unknown kind visible.

---

## Doc bug: module docstring says `storage`, code uses `collection`

**`server/content-types.mjs`**, line 9:

> "each descriptor's `id`, `displayName`, `storage`, and ordered `fields`"

The field is `collection: string | null`, not `storage`. `storage` doesn't appear anywhere in the typedef or data. Should read:

> "each descriptor's `id`, `displayName`, `collection`, and ordered `fields`"

---

## Consistency: `createTyped` skips the path-safety guard that `createPost` has

**`server/create-content.mjs`**:

`createPost` explicitly validates the caller-supplied `collection` against `/^[A-Za-z0-9_-]+$/` before using it as a path segment. `createTyped` uses `descriptor.collection` directly — hardcoded, so the blast radius is zero today, but the pattern is inconsistent and harder to audit as new types are added. A one-liner assertion keeps the contract uniform:

```js
if (!/^[A-Za-z0-9_-]+$/.test(collection)) {
  throw new Error(`Internal error: invalid collection name in registry: ${collection}`);
}
```

---

## Test gap: no path-traversal test for `createTyped`

**`test/create-content.test.js`**:

The `createPost` suite has a path-traversal test. `createTyped` doesn't. `slugify` makes this safe already, but a test locks down the property:

```js
it("title with path separators cannot escape the collection dir", () => {
  const result = createTyped(repo, { type: "note", title: "../../../etc/evil" });
  expect(existsSync(join(repo, "..", "..", "..", "etc", "evil.md"))).toBe(false);
  expect(result.filePath).toMatch(/^src\/content\/notes\//);
});
```


---

davidwkeith:
Thanks — all five addressed in 8c0f486:

1. **`businessProfile` schema cliff** — added `creatableContentTypeIds` (collection-only) to `content-types.mjs`; `create_content` now uses it, so the schema no longer advertises page-stored types. The `createTyped` guard stays as defence-in-depth.
2. **`renderEntry` silently drops unknown kinds** — added a `default` case that throws `renderEntry: unhandled field kind "…"`; `createTyped` catches and surfaces it.
3. **Docstring `storage` → `collection`** — fixed.
4. **`createTyped` path-safety** — added the same `/^[A-Za-z0-9_-]+$/` assertion `createPost` uses against `descriptor.collection`.
5. **Path-traversal test** — added (title `../../../etc/evil` stays inside `src/content/notes/`).

Suite green: `create-content` 19 passed, `mcp-server` 8 passed.

Issue #320, opened 2026-05-27
template: npm run ai-optimize broken — '@dwk/anglesite/server/optimize-images.mjs' doesn't resolve from inside the package


davidwkeith:
Surfaced during the Phase 9 step 3 image-drop smoke (see [Anglesite/Anglesite-app#32](https://github.com/Anglesite/Anglesite-app/issues/32) and #319 which just merged).

## The problem

`template/scripts/optimize-images.ts` was thinned in #319 to a CLI wrapper around the new `server/optimize-images.mjs`. The wrapper does:

```ts
const pluginServerOptimize = "@dwk/anglesite/server/optimize-images.mjs";
const { optimizeImage } = await import(pluginServerOptimize);
```

This works when `@dwk/anglesite` is installed *as a dependency* (e.g. a hypothetical future site that depends on the plugin). But `template/package.json` declares the template itself as `"name": "@dwk/anglesite"`, and sites scaffolded from the template inherit that name (a site IS the package, not a consumer of it). So:

```
~/Sites/<name>/scripts/optimize-images.ts
                 └─ import "@dwk/anglesite/server/optimize-images.mjs"
                    └─ node module resolution looks for `node_modules/@dwk/anglesite/...`
                       └─ doesn't exist; @dwk/anglesite isn't in `node_modules/`
                          └─ ERR_MODULE_NOT_FOUND
```

## Repro

```sh
cd ~/Sites/anglesite-smoke           # any site scaffolded from the template
npm run ai-optimize
# → Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@dwk/anglesite' imported from .../scripts/optimize-images.ts
```

## Impact

- `npm run ai-optimize` CLI is broken on every existing scaffolded site once they rsync the new template script.
- The image-drop pipeline (PR #319, run via the plugin's MCP server) is **unaffected** — the server imports the local `./optimize-images.mjs` via a relative path, which works because the dispatcher and the optimize core live in the same `server/` directory.

## Suggested fix

The site IS `@dwk/anglesite`, so the CLI script can use a relative path to its own server module:

```ts
// template/scripts/optimize-images.ts
import { fileURLToPath } from "node:url";
import { dirname, resolve } from "node:path";
const here = dirname(fileURLToPath(import.meta.url));      // .../scripts
const pluginRoot = resolve(here, "..");                    // site root === plugin root
const { optimizeImage } = await import(`${pluginRoot}/server/optimize-images.mjs`);
```

[...]

PR #391, opened 2026-07-01
fix: clear prod npm audit advisories + make the #320 optimize e2e actually run

davidwkeith:
## Summary

Two review-midpoint fixes: production dependency advisories that ship inside the Mac app bundle, and a #320 regression e2e that could never pass anywhere.

### 1. Production npm audit advisories (lockfile-only)

`npm audit --omit=dev` at HEAD reported **6 vulnerabilities (2 high, 4 moderate)**, all transitive under `@modelcontextprotocol/sdk` — these ship in the app bundle's production `node_modules`:

| Package | Severity | Before → After |
|---|---|---|
| hono | high | 4.12.9 → 4.12.27 |
| fast-uri | high | 3.1.0 → 3.1.3 |
| @hono/node-server | moderate | 1.19.11 → 1.19.14 |
| qs | moderate | 6.15.0 → 6.15.3 |
| ip-address (via express-rate-limit) | moderate | 10.1.0 → 10.2.0 (8.3.1 → 8.5.2) |

Fixed with plain `npm audit fix` (no `--force`): all bumps are in-range, `package.json` is untouched, and the SDK stays at 1.28.0. The same pass clears the **critical** dev-only vitest advisory (GHSA-5xrq-8626-4rwp; vitest → 3.2.6, vite → 7.3.6). The remaining dev-only advisories are all under `markdownlint-cli2` and require a breaking 0.23 bump, left for a deliberate upgrade.

### 2. #320 optimize e2e never passed anywhere

`test/optimize-images-packaging.test.js`'s "runs the optimize script against a fixture JPEG without ERR_MODULE_NOT_FOUND" scaffolds a site into a temp dir **without** npm-installing it, then runs the CLI there. `server/optimize-images.mjs` does `import("sharp")`, which a real site resolves from its **own** `node_modules` (sharp is a template devDependency) — the uninstalled temp scaffold has neither that nor a walk-up path to the repo's, so the test failed with `ERR_MODULE_NOT_FOUND` on every macOS dev machine. On CI the file is `describe.skipIf(!hasZsh)` and Ubuntu runners lack zsh, so it silently skipped. Net: the regression guard guarded nothing.

**Fix:** after scaffolding, symlink the repo's `node_modules` into the site root as a stand-in for the site's `npm install` — the same first-walk-up-hop resolution a real installed site uses, carrying the same `sharp ^0.33` the template pins. This does **not** weaken #320's guard: `node_modules/@dwk` doesn't exist in the repo install, so a regressed bare `@dwk/anglesite` import still fails through the link. Verified by temporarily reintroducing the bare specifier: both the e2e and the static import-target guard failed as expected, then reverted.

## Verification

- `npm audit --omit=dev` → **0 vulnerabilities**
- `npx vitest run test/optimize-images-packaging.test.js` → 5/5 pass; the e2e now exercises the full sharp pipeline (WebP variant emitted, original preserved in `originals/`)
- `npm test` → **125 files, 2787 tests, all passing** (on the bumped vitest 3.2.6)

Refs #320.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,6 +24,13 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
+      # The optimize-images packaging e2e (test/optimize-images-packaging.test.js, #320)
+      # scaffolds a real site via scaffold.sh, which needs zsh (and rsync) — both absent
+      # from ubuntu-latest, so the suite's describe.skipIf(!hasZsh) silently skipped it
+      # here and the regression guard only ever ran on contributor Macs.
+      - name: Install zsh + rsync (for the scaffold e2e)
+        run: sudo apt-get update -q && sudo apt-get install -y --no-install-recommends zsh rsync
+
       - run: npm ci
 
       - name: Verify skill registry is up to date
--- a/package-lock.json
+++ b/package-lock.json
@@ -703,9 +703,9 @@
       }
     },
     "node_modules/@hono/node-server": {
-      "version": "1.19.11",
-      "resolved": "https://registry.npmjs.org/@hono/node-server/-/node-server-1.19.11.tgz",
-      "integrity": "sha512-dr8/3zEaB+p0D2n/IUrlPF1HZm586qgJNXK1a9fhg/PzdtkK7Ksd5l312tJX2yBuALqDYBlG20QEbayqPyxn+g==",
+      "version": "1.19.14",
+      "resolved": "https://registry.npmjs.org/@hono/node-server/-/node-server-1.19.14.tgz",
+      "integrity": "sha512-GwtvgtXxnWsucXvbQXkRgqksiH2Qed37H9xHZocE5sA3N8O8O8/8FA3uclQXxXVzc9XBZuEOMK7+r02FmSpHtw==",
       "license": "MIT",
       "engines": {
         "node": ">=18.14.1"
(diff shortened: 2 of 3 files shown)


davidwkeith (commented):
Reviewed both changes. I checked out this branch locally, ran `npm ci` + `npm audit --omit=dev` (confirms 0 vulnerabilities as claimed), ran the full suite (2787 tests — only 2 unrelated failures in `apply-edit-dispatcher.test.js`, caused by running as root in my sandbox so a "directory isn't writable" permission check doesn't actually fail; unrelated to this diff), and specifically exercised the fixed e2e test after installing `zsh`/`rsync` locally — it passes and produces the expected WebP output. I also empirically verified `rmSync(tmpDir, { recursive: true, force: true })` in `afterEach` does **not** follow the new `node_modules` symlink into the real repo `node_modules` (Node unlinks the symlink itself rather than recursing into its target) — so the cleanup is safe and won't ever delete the plugin's real `node_modules`.

One thing worth fixing before this really closes out #320:

**The e2e test still won't run in CI.** `.github/workflows/test.yml`'s `test` job runs on `ubuntu-latest` and never installs `zsh`. The PR body itself notes "Ubuntu runners lack zsh, so it silently skipped" — but this PR doesn't add a zsh install step, so after merge the `describe.skipIf(!hasZsh)` block will keep silently skipping in CI exactly as before. I reproduced this: running `npx vitest run test/optimize-images-packaging.test.js` in an unmodified Ubuntu environment reports `2 skipped`; only after `apt-get install zsh rsync` did the two e2e tests actually execute. The fix to the symlink/resolution logic is correct and does make the test runnable, but as merged the "actually run" part of the PR title is only true on a contributor's local machine (e.g. macOS, which ships zsh), not in automated CI. Consider adding a `zsh` install step to the `test` job (or documenting that this remains a local-only regression guard) so #320 has an actual CI backstop.

Minor/non-blocking: the new symlink trick's safety depends on `node_modules/@dwk` not existing in the plugin's own root install (confirmed true today — no `@dwk/*` scoped dependency in root `package.json`). If a real `@dwk/*` package (e.g. the indieauth/micropub/webmention packages referenced in `vitest.config.ts`) is ever added as a root dependency, this guard would silently stop catching a regressed bare `@dwk/anglesite` import, since it'd then resolve through the symlink. Not urgent, but worth a one-line comment or an explicit assertion (`expect(existsSync(join(tmpDir, 'node_modules', '@dwk'))).toBe(false)`) so it fails loudly instead of silently if that invariant ever breaks.

Everything else — the audit-fix lockfile bumps, the byte-identical-copy guard, the CLI import-target assertion — checks out.

---

davidwkeith:
Both review points addressed in 3d25315:

- **CI backstop**: the `test` job now installs `zsh` + `rsync` on ubuntu-latest before running the suite, so the #320 scaffold e2e actually executes in CI instead of silently skipping via `describe.skipIf(!hasZsh)` — the matrix run on this PR is the proof.
- **Symlink invariant**: added the explicit assertion that `node_modules/@dwk` doesn't exist at the repo root, with a comment explaining why — if a `@dwk/*` package is ever installed there, the guard now fails loudly instead of silently resolving a regressed bare import through the link.

Verified locally: `npx vitest run test/optimize-images-packaging.test.js` → 5/5 pass with the new assertion.

davidwkeith (commented):
Reviewed and verified independently:

- `npm audit --omit=dev` on this branch → 0 vulnerabilities, matches the PR description.
- The CI workflow change correctly scopes the `zsh`/`rsync` install to the `test` job only (before `npm ci`), not the other jobs that don't need it.
- `test/optimize-images-packaging.test.js` passes here (e2e steps skip in this sandbox since it also lacks zsh, same failure mode the PR is fixing on CI — consistent with the described bug).
- The symlink-into-tmpDir approach is sound: Node's recursive `rmSync` unlinks symlinks rather than recursing through them, so `afterEach`'s cleanup won't touch the real repo `node_modules`.

No issues found. Looks good to merge.

---

PR #395, opened 2026-07-02
feat(indieweb): close h-feed, h-event, and webmention-sending gaps

davidwkeith:
## Summary

- Blog listing page was missing the `h-feed` wrapper `/anglesite:check`'s own checklist already expects — every scaffolded site was failing its own audit. Wrapped `template/src/pages/blog/index.astro` to match the pattern already shipped for `/notes`.
- The `events` Keystatic collection had a content schema but no pages — h-event markup was documentation-only. Added `/events/index.astro` and `/events/[slug].astro` rendering `h-event` microformats + `Event` JSON-LD, following the podcast episode page pattern (`jsonLd` prop on `BaseLayout`).
- Webmention was receive-only. Added `template/scripts/send-webmentions.ts`: scans built posts/notes for outbound links via `microformats-parser` (already a dependency once Webmention is enabled), discovers each target's webmention endpoint (`Link` header, then `<link rel="webmention">`), and sends. Wired into `/anglesite:deploy` behind `INDIEWEB_WEBMENTION=true`. Includes a DNS-resolved SSRF guard (blocks private/loopback targets before any request) and a `webmention-sent.json` ledger committed to the repo so a given `(post, link)` pair is attempted once, ever.
- Documented Brid.gy as an optional backfeed path (likes/replies/reposts on syndicated social copies flowing back as webmentions) — framed explicitly as the one third-party piece in an otherwise self-owned stack, not a default step.
- Corrected two stale "`@dwk/*` packages unpublished" notes in ADR-0020 and `template/docs/indieweb.md` — confirmed via `npm view` that they're live at `0.1.0-beta.2`+.

## Paired PR check

- [x] This change is **self-contained** to the plugin (skills / template / docs, no MCP message schema or hook surface change).

## Test plan

- [x] `npm test` — 2812/2813 passing (the one failure is a pre-existing missing `tsx` binary in `node_modules/.bin`, unrelated to this change and reproducible on `main`)
- [x] `npm run build:agent-skills` — regenerated, no longer stale
- [x] `scripts/scaffold.sh`-equivalent (`rsync` copy of `template/`) produces a buildable site; verified via `astro dev` that:
  - `/blog/` renders `<div class="h-feed">` wrapping `h-entry` items
  - `/events/` and `/events/[slug]/` render `h-event` microformats and valid `Event` JSON-LD
  - `send-webmentions.ts` runs end-to-end against real external domains (example.com, iana.org), correctly reports "no endpoint" for both, and is idempotent on re-run (skips already-attempted pairs)
- [x] New test file `tests/send-webmentions.test.ts` — 20 tests covering link extraction, endpoint discovery, the SSRF guard, sending, ledger persistence, and full-site orchestration

### DIFF
--- a/template/scripts/send-webmentions.ts
+++ b/template/scripts/send-webmentions.ts
@@ -1,0 +1,328 @@
+/**
+ * Build-time webmention sender.
+ *
+ * ADR-0020 composes a receiving @dwk/webmention endpoint into the site
+ * Worker, but a site never sent webmentions for its own outbound links —
+ * the one spot where the self-owned IndieWeb story fell back to "use a
+ * third-party tool" (template/docs/indieweb.md). This script closes that
+ * gap at deploy time: it scans built h-entry pages (blog posts, notes) for
+ * external links in their e-content, discovers each target's webmention
+ * endpoint, and sends one.
+ *
+ * Runs against `dist/` after `astro build`, not against source Markdoc —
+ * `microformats-parser` (already a dependency once Webmention is enabled;
+ * see worker/webmention-inbox.js) scopes extraction to the actual h-entry
+ * content instead of nav/footer boilerplate that a raw regex would also
+ * catch.
+ *
+ * A (source, target) pair is attempted at most once, ever — tracked in a
+ * ledger file committed to the repo (same precedent as `.site-config`:
+ * cross-deploy state lives in a project file, not a Cloudflare binding).
+ * This bounds the number of outbound network calls per deploy and avoids
+ * hammering targets that don't support webmentions.
+ *
+ * Usage: tsx scripts/send-webmentions.ts [--dist dist] [--ledger webmention-sent.json]
+ *
+ * @module
+ */
+
+import { existsSync, mkdirSync, readdirSync, readFileSync, writeFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import dns from "node:dns";
+import { mf2 } from "microformats-parser";
+import { extractLinks, isExternalLink } from "./link-check.js";
+import { readConfig } from "./config.js";
+
+// ---------------------------------------------------------------------------
+// Outbound link extraction
+// ---------------------------------------------------------------------------
+
+/**
+ * Extracts external links from the `e-content` of the first `h-entry` in
+ * `html`, resolved to absolute URLs and deduped. Returns `[]` if there's no
+ * h-entry or e-content — links in the nav, footer, or other page chrome are
+ * never included even though they're on the same page.
+ */
+export function extractOutboundLinksFromEntry(html: string, permalink: string): string[] {
+  let items;
+  try {
+    ({ items } = mf2(html, { baseUrl: permalink }));
+  } catch {
+    return [];
+  }
+
+  const entry = items.find((item: { type?: string[] }) => item.type?.includes("h-entry"));
+  const contentHtml: string | undefined =
+    entry?.properties?.content?.[0]?.html ?? entry?.properties?.content?.[0]?.value;
+  if (!contentHtml) return [];
+
+  const siteOrigin = new URL(permalink).origin;
+  const seen = new Set<string>();
+  const links: string[] = [];
+  for (const href of extractLinks(contentHtml)) {
+    let absolute: string;
+    try {
+      absolute = new URL(href, permalink).toString();
+    } catch {
+      continue;
+    }
+    if (!isExternalLink(absolute)) continue;
+    if (new URL(absolute).origin === siteOrigin) continue;
+    if (seen.has(absolute)) continue;
+    seen.add(absolute);
+    links.push(absolute);
+  }
+  return links;
+}
+
+// ---------------------------------------------------------------------------
+// Webmention endpoint discovery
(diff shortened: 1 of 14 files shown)


template/scripts/send-webmentions.ts
```diff
@@ -0,0 +1,328 @@
...
+      }
+
+      const endpoint = await discoverWebmentionEndpoint(target, fetchImpl);
+      if (!endpoint) {
+        ledger[key] = { status: "no-endpoint", at: new Date().toISOString() };
+        summary.noEndpoint++;
+        continue;
+      }
+
+      const result = await sendWebmention(endpoint, permalink, target, fetchImpl);
```
davidwkeith, line 286: **SSRF: the discovered endpoint is never passed through `isSafeUrl`.**

`isSafeUrl` (line 272) is only checked against `target` — the outbound link found in the owner's own post. `endpoint` (line 279) is parsed out of *that target's own HTTP response* (the `Link` header or an `<link rel=webmention>`/`<a rel=webmention>` tag), i.e. it's attacker-controlled data coming back from whatever site the owner linked to. It's then POSTed to directly here with no safety check at all.

Concretely: if `target` is (or becomes) a page that advertises `Link: <http://169.254.169.254/latest/meta-data/>; rel="webmention"` (or any other internal/loopback URL), this script will POST to it from whatever machine/CI runner executes `/anglesite:deploy` — the exact class of request the SSRF guard exists to prevent, just one hop later.

Fix: run `endpoint` through `isSafeUrl(endpoint, lookupImpl)` (same as `target`) before calling `sendWebmention`, and treat an unsafe endpoint the same as `blocked`/`failed`.

---

template/scripts/send-webmentions.ts
```diff
@@ -0,0 +1,328 @@
...
+  /<(?:link|a)\b[^>]*\bhref\s*=\s*["']([^"']+)["'][^>]*\brel\s*=\s*["'][^"']*\bwebmention\b[^"']*["'][^>]*>/i;
+
+/** Follows the webmention discovery algorithm: HTTP Link header, then HTML `<link>`/`<a rel=webmention>`. */
+export async function discoverWebmentionEndpoint(
+  targetUrl: string,
+  fetchImpl: typeof fetch = fetch,
+): Promise<string | null> {
+  let response: Response;
+  try {
+    response = await fetchImpl(targetUrl, {
```
davidwkeith, line 96: Secondary, lower-priority gap in the same guard: `isSafeUrl` (line 150) resolves the hostname once via its own `dns.promises.lookup`, but the actual request here uses the real `fetch`, which does its own independent DNS resolution (and follows redirects by default). A target that resolves to a public IP at check-time and a private/loopback IP at request-time (DNS rebinding), or that 30x-redirects to an internal URL, bypasses the check entirely — nothing pins the fetch to the address that was validated. Not blocking given the primary endpoint-validation gap is the more direct exploit path, but worth a follow-up if this guard is meant to be relied on (e.g. resolve once, pass `redirect: "manual"`, and validate each hop / connect to the checked IP directly).

---

template/src/pages/events/[slug].astro
```diff
@@ -0,0 +1,85 @@
...
+// the human-readable time is shown alongside it, not folded into the
+// datetime attribute.
+const isoDate = event.data.date.toISOString();
+
+const eventJsonLd = {
+  "@context": "https://schema.org",
+  "@type": "Event",
+  name: event.data.title,
+  startDate: isoDate,
+  ...(event.data.endTime && { endDate: isoDate }),
```
Comment (davidwkeith) [inline template/src/pages/events/[slug].astro:30]: Minor: when `endTime` is set, `endDate` is emitted as `isoDate` — the exact same value as `startDate` above (both are the day-level date; `endTime` is free text and never makes it into the datetime value). So any event with an end time gets `startDate === endDate` in the JSON-LD, which reads as a zero-duration event and doesn't actually encode the fact that there's a distinct end time. Since it carries no more information than omitting `endDate` entirely, and a search engine parsing this literally sees "starts and ends at the same instant," it'd probably be better to leave `endDate` out altogether (matching what's done for `dt-end` being an honest reflection would require timezone-aware start/end times, which the schema doesn't have) rather than emit a value that looks precise but isn't. Same mismatch applies to the `dt-end` microforment attribute a few lines below.

---

davidwkeith (commented):
Nice feature set (h-feed wrapper, real h-event pages, outbound webmention sending), and I verified the non-security parts: `npm run build:agent-skills` is not stale on this branch, and `tests/send-webmentions.test.ts` (20 tests) passes.

One finding I'd treat as blocking: the outbound webmention sender's SSRF guard (`isSafeUrl`) validates the *target* link but never validates the *discovered endpoint* before POSTing to it, and that endpoint is parsed straight out of the target's own (attacker-controlled) HTTP response. See the inline comment on `template/scripts/send-webmentions.ts` for the concrete fix. Two lower-priority follow-ups also inline (DNS-rebinding/redirect TOCTOU on the same guard, and a `startDate === endDate` JSON-LD nit on the event page) — neither blocking on their own.

(Note: I couldn't mark this "Request changes" — GitHub doesn't allow requesting changes on your own PR under this token's identity — so treat the SSRF finding above as the actionable item instead.)

---

PR #397, opened 2026-07-04
feat(design-import): Safari MCP rendered-extraction backend for Canva import

davidwkeith:
## Summary

- Migrates the design-import skill to the Safari MCP rendered-extraction backend (`safaridriver --mcp`, STP 247+): new `scripts/design-import/canva-safari.mjs` runs the existing Canva page-functions and site crawl **unmodified** over Safari via a four-method Playwright-`page` adapter, emitting the exact same JSON contract as `canva-playwright.mjs`, plus `--check` with the shared exit codes (0 usable / 2 not installed / 3 not enabled / 4 session failure).
- SKILL.md Step 1 becomes "Choose a browser backend" — Safari preferred on macOS (no 150 MB Playwright install), exit-code branching with one-time enable instructions, silent Playwright fallback on Linux/Windows — mirroring the import skill's planned Step 1a.2 (ADR-0021 accelerator pattern; ADR-0024 pending in the base work).
- Assessment recorded in `docs/superpowers/specs/2026-07-03-design-import-safari-backend-design.md`: the authenticated-session caveat is **not a blocker** (the skill only accepts published `*.my.canva.site` URLs, and the Playwright driver already runs with a cookie-free fresh browser); Figma has no extraction code to migrate.

**Includes the base Safari infrastructure commits** (`feecd91`, `402edd0`, `9eefb73` — Tasks 1–3 of the `2026-07-03-safari-import-backend` plan from `claude/blissful-mendeleev-a2581c`): the stdio JSON-RPC client, shared wix/squarespace page-functions, the `extractContentSrc` ReferenceError fix, and the fake-safaridriver CI fixture. If that branch PRs separately, land it first and this diff shrinks to the design-import commits (`601f3b3`, `8e5fbf8`).

Also freshens the `agent-skills/` export (the base commits had left it stale — CI's `agent-skills-export` job would have failed) and documents a pre-existing exporter gap: bundled scripts' relative JS imports aren't copied (fix in progress as a separate task).

## Paired PR check

- [ ] This change **needs a paired PR** in [`Anglesite/Anglesite-app`](https://github.com/Anglesite/Anglesite-app)

## Test plan

- [x] Full suite green: 129 files / 2836 tests, including 7 new tests driving the CLI end-to-end against the fake safaridriver (`--check` exit codes, single-page token/section/nav/image contract through the real color+font pipeline, `--content-only`, `--site` crawl with homepage-only tokens and image dedup, all-failed exit 1)
- [x] `agent-skills/` export regenerated and fresh (`npm run build:agent-skills`)
- [x] Live `--check` validated against real Safari Technology Preview on macOS (exit 0, session created)
- [ ] Manual before release: live `--site` extraction against a real published Canva URL, diffed against the Playwright driver's output


scripts/import/browser/safari-mcp.mjs
```diff
@@ -0,0 +1,127 @@
...
+      const timer = setTimeout(() => {
+        if (this.pending.has(id)) {
+          this.pending.delete(id);
+          reject(new SafariMcpError('timeout', `${method} timed out after ${timeoutMs}ms`));
+        }
+      }, timeoutMs);
+      timer.unref?.();
+      this.pending.set(id, { resolve, reject });
+    });
+    this.child.stdin.write(JSON.stringify({ jsonrpc: '2.0', id, method, params }) + '\n');
```
davidwkeith, line 125: **Unhandled `stdin` write errors will crash the process.** `this.child.stdin.write(...)` is called here and at line 82 with no `'error'` listener on the `stdin` stream itself. The `this.child.on('error', ...)` handler at line 59 only covers spawn-level failures (e.g. ENOENT) — it does not fire for an EPIPE from writing to a stream whose reader has gone away.

Since safaridriver's `--mcp` mode opens a visible Safari window per this file's own comments, a user closing that window (or the process crashing/being killed) mid-session leaves `this.child.stdin` writable-but-broken. The next `#request()` call throws an unhandled `'error'` event on the stream, which crashes the whole import/design-import process instead of surfacing a catchable `SafariMcpError`.

Suggest adding `this.child.stdin.on('error', () => {})` (or routing it through the same pending-rejection path as the `child.on('error', ...)` handler) in `start()`.

---

scripts/import/browser/safari-mcp.mjs
```diff
@@ -0,0 +1,127 @@
...
+    this.pending = new Map();
+    this.nextId = 1;
+  }
+
+  async start() {
+    const args = ['--mcp'];
+    this.child = this.binaryPath.endsWith('.mjs')
+      ? spawn(process.execPath, [this.binaryPath, ...args], { stdio: ['pipe', 'pipe', 'ignore'] })
+      : spawn(this.binaryPath, args, { stdio: ['pipe', 'pipe', 'ignore'] });
+    this.child.on('error', (err) => {
```
davidwkeith, line 69: No `'exit'`/`'close'` handler is registered on `this.child`. If safaridriver dies unexpectedly while requests are pending (crash, killed, window closed), nothing rejects the in-flight promises immediately — each just waits out its own timer (up to 60000ms for `call()`, 15000ms for `initialize()`). During a `--site` crawl that's a long, opaque hang instead of a fast, actionable error. Worth rejecting all `this.pending` entries (same as the `'error'` handler above) when the child exits unexpectedly.

---

test/fixtures/fake-safaridriver.mjs
davidwkeith, line 1: Test coverage gap: the PR description claims a 0/2/3/4 exit-code contract for `canva-safari.mjs --check`, but only exit codes 0, 2, and 3 are actually exercised in `test/canva-safari.test.js`. This fixture only implements `ok` / `not-enabled` / `hang` modes — none of them drives `mcp.start()` to fail or a `tools/call` to return a generic (non-`not-enabled`) error, so `EXIT.SESSION_FAILED` (4) is reachable in the real code but never verified by CI. Consider adding a `fail`-style mode here (e.g. a bogus `tools/call` error response, or an immediate non-zero exit) and a corresponding test case.

Related: `waitForSelector`'s deadline-exceeded branch in `canva-safari.mjs` (throws `SafariMcpError('page-failure', 'Timed out waiting for selector...')`) is also dead code in CI — this fixture's `querySelectorAll(...).length` check always answers `"1"` on the first poll, so the timeout path never runs.

---

davidwkeith (commented):
Reviewed the Safari MCP backend migration. Core logic is solid — the 4-method Canva adapter correctly covers what `canva-playwright.mjs`'s extraction code calls, and the wix-playwright.mjs refactor into `page-functions.mjs` is a faithful extraction (it even incidentally removes a pre-existing dead-code bug in the old "full-page mode" branch). Left two correctness comments on `safari-mcp.mjs` around unhandled stream/process failure modes, plus a test-coverage gap on the claimed exit-code contract.

---

davidwkeith:
Addressed the review feedback in eba2caa:

1. **stdin EPIPE crash** — added `this.child.stdin.on('error', ...)` in `SafariMcp.start()`. Previously an EPIPE from writing to a stdin stream whose reader has gone away (e.g. Safari window closed mid-session) had no listener and would crash the whole process. It now routes through the same `SafariMcpError('session-failed', ...)` path as the child's own `'error'` handler.

2. **Silent hang on unexpected exit** — added an `'exit'` handler on `this.child` that immediately rejects every entry in `this.pending` with a `session-failed` error and sets a new `#dead` flag. `#request()` now checks `#dead` up front and fails fast for any calls made after the process has died, instead of letting each in-flight call wait out its own timeout (up to 60s for `call()`, 15s for `initialize()`). Refactored the shared "reject everything in `pending`" logic into a small `#rejectAllPending()` private helper used by both the `'error'` and `'exit'` handlers.

3. **Exit code 4 coverage (nice-to-have)** — added a `fail` mode to `test/fixtures/fake-safaridriver.mjs` that returns a generic (non-`not-enabled`) JSON-RPC tool error, plus a new test in `test/canva-safari.test.js` (`exits 4 on a generic session failure (not the not-enabled case)`) asserting `canva-safari.mjs --check` exits with `EXIT.SESSION_FAILED` (4) in that case. This was implemented — not skipped.

**Test results:** `test/canva-safari.test.js` (8 tests, was 7) and `test/safari-mcp.test.js` (5 tests) all pass. Full `npm test`: 2831/2832 passing — the one failure (`test/optimize-images-packaging.test.js`, missing `tsx` binary in this sandbox) is pre-existing and unrelated to this change (reproduced identically on the pre-fix commit).

PR #398, opened 2026-07-04
fix(agent-skills): bundle the relative-import closure of copied scripts

davidwkeith:
## Summary

- `bin/build-agent-skills.ts` copied scripts referenced from SKILL.md `${CLAUDE_PLUGIN_ROOT}` paths into `references/` but ignored the scripts' own relative `import` statements, so bundled scripts crashed on import in a standalone agent-skills install (e.g. `design-import`'s `canva-playwright.mjs` imports `./canva-colors.mjs` / `./canva-fonts.mjs`, `import`'s `wix-playwright.mjs` imports `./color-utils.mjs` — none were bundled).
- The emitter now statically walks each copied script's relative import graph (`import`/`export … from`, side-effect, and dynamic imports) and copies the transitive closure into `references/` at plugin-root-relative paths, so imports resolve unchanged. It also resolves the TS ESM convention (`./x.js` compiled from `./x.ts`), which additionally fixes `booking`'s bundled `csp.ts` → `config.ts`. Unresolvable imports surface as `MISSING IMPORT` build warnings instead of silently broken bundles.
- Adds unit tests for the closure walk (re-export chains, parent-directory imports, directory bundles, missing-import warning) plus a sync guard asserting every committed bundled script's relative imports resolve; regenerates `agent-skills/` (5 new bundled files). Documents the behavior in `docs/dev/agent-skills.md`.

## Paired PR check

## Test plan

- [x] `npm test` — 126 files, 2832 tests passing (includes the new closure tests and the `agent-skills` sync guards)
- [x] `npm run build:agent-skills` re-run on top of the commit produces zero diff, so CI's `agent-skills-export` job stays green
- [ ] If touching the template: `scripts/scaffold.sh` produces a buildable site (template not touched)
- [ ] If touching the MCP server: paired app PR exercises the new/changed messages (MCP server not touched)

## Reviewer notes

- The bug report also mentioned `canva-safari.mjs` and `../browser/page-functions.mjs`; those files don't exist on `main`/this branch (likely from unlanded work). When they land, the closure walk picks them up automatically and the new sync-guard test enforces it.
- Likewise, the `docs/dev/agent-skills.md` "Relative JS imports are not followed" limitation bullet to remove didn't exist here; the doc now documents the closure-walking behavior in "What the transformer does" instead.

### DIFF
--- a/tests/build-agent-skills.test.ts
+++ b/tests/build-agent-skills.test.ts
@@ -201,6 +201,89 @@ describe("emitSkill", () => {
     const warnings = emitSkill(result("demo", ["scripts/does-not-exist.mjs"]), pluginRoot, outRoot);
     expect(warnings.some((w) => w.startsWith("MISSING REFERENCE:") && w.includes("does-not-exist"))).toBe(true);
   });
+
+  it("bundles the transitive relative-import closure of a copied script", () => {
+    writePlugin(
+      "scripts/a/entry.mjs",
+      "import { x } from './sibling.mjs';\nimport { y } from '../b/dep.mjs';\nexport const z = x + y;\n",
+    );
+    writePlugin("scripts/a/sibling.mjs", "export { s } from './deep.mjs';\nexport const x = 1;\n");
+    writePlugin("scripts/a/deep.mjs", "export const s = 2;\n");
+    writePlugin("scripts/b/dep.mjs", "export const y = 3;\n");
+    const warnings = emitSkill(result("demo", ["scripts/a/entry.mjs"]), pluginRoot, outRoot);
+    // Direct imports, a re-export chain, and a parent-directory import all land
+    // in references/ at their plugin-root-relative paths.
+    expect(existsSync(join(outRoot, "demo", "references/scripts/a/sibling.mjs"))).toBe(true);
+    expect(existsSync(join(outRoot, "demo", "references/scripts/a/deep.mjs"))).toBe(true);
+    expect(existsSync(join(outRoot, "demo", "references/scripts/b/dep.mjs"))).toBe(true);
+    expect(warnings).toEqual([]);
+  });
+
+  it("warns when a copied script imports a relative path that does not exist", () => {
+    writePlugin("scripts/a/entry.mjs", "import { x } from './gone.mjs';\n");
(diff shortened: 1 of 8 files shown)


tests/build-agent-skills.test.ts
```diff
@@ -201,6 +201,89 @@ describe("emitSkill", () => {
...
+});
+
+// ---------------------------------------------------------------------------
+// Bundled scripts must be runnable: every relative import inside a bundled
+// .mjs/.js/.ts file must resolve within the same references/ tree.
+// ---------------------------------------------------------------------------
+
+describe("bundled scripts resolve their relative imports", () => {
+  const SCRIPT_EXT_RE = /\.(mjs|cjs|js|ts|mts)$/;
+  const IMPORT_RES = [
```
davidwkeith: This hand-copies the three relative-import regexes instead of importing and reusing the `relativeImports()` function that `bin/build-agent-skills.ts` already exports (added a few lines above in this same PR, `bin/build-agent-skills.ts:283`). 

That undermines the point of this being a "sync guard": if the real regex logic in `relativeImports()` is later tightened or loosened (e.g. to handle a new import syntax, or to fix a false-positive match inside a comment), this test's copy won't change with it, so it will keep asserting against stale matching behavior rather than the actual implementation it's meant to guard. Suggest importing `relativeImports` from `../bin/build-agent-skills.ts` here (it's already imported at the top of the file) and dropping the local `IMPORT_RES` array.

---

davidwkeith (commented):
Reviewed the relative-import closure walker. The core `bin/build-agent-skills.ts` logic looks correct — cycle-safe via the `bundled` set guard in `copyInto`, correctly handles re-export chains and parent-directory imports, and only matches genuinely relative specifiers (bare package imports are correctly skipped). One maintainability finding on the new test file below.

---

tests/build-agent-skills.test.ts
davidwkeith: Fixed in f90a521: the test now imports `relativeImports` from `../bin/build-agent-skills.ts` and calls it directly instead of maintaining a separate `IMPORT_RES` array. Ran `tests/build-agent-skills.test.ts` (86/86 passing) and the full suite (2831/2832 passing — the one failure is a pre-existing, unrelated missing-`tsx`-binary issue in this sandbox).

PR #399, opened 2026-07-04
Prototype: network-layer Wix blog extraction via Safari MCP (GO for posts, NO-GO for static pages)

davidwkeith:
## Summary

- Prototypes the follow-up flagged in `docs/superpowers/specs/2026-07-03-safari-import-backend-design.md`: recovering Wix content from captured network JSON (Safari MCP `list_network_requests`/`get_network_request`) instead of DOM scraping. Verdict, validated against five live pages: **GO for blog posts, NO-GO for static pages** — full comparison and ADR open questions in `docs/superpowers/specs/2026-07-03-wix-network-extraction-findings.md`.
- Adds `scripts/import/browser/wix-network.mjs`, the draft `--network` mode: handles both live Wix blog delivery generations (adapter-API XHR with resolved categories/owner, and SSR `wix-warmup-data` + JSON-LD), converts Ricos richContent to Markdown (GFM tables, collapsibles without click-expand, original-resolution image media IDs, video URLs), emits the NDJSON driver contract, and returns `post: null` for non-blog pages as the DOM-fallback signal. Includes a `--from-file` offline mode for fixture tests.
- **Exploratory — deliberately not wired into the import skill**; adoption needs a new ADR. `safari-mcp.mjs` and `page-functions.mjs` are copies of in-flight work from `claude/blissful-mendeleev-a2581c` (where `safari-driver.mjs` will land) so the module runs standalone; the two branches need reconciling at merge.

## Paired PR check

## Test plan

- [x] `node --check` on the new module; full suite run — 2818/2819 pass, the one failure (`test/optimize-images-packaging.test.js` 10 s timeout) reproduces identically on untouched `main` (pre-existing, flagged separately)
- [x] Live validation (macOS + Safari Technology Preview 247, remote automation on): three wix.com/blog posts (SSR-warmup generation), one template-demo post (adapter-API generation), one static demo page (correctly yields `null` → DOM fallback); complete 2.45–3.6 MB response bodies captured untruncated
- [ ] Not applicable: template and MCP server untouched

## Reviewer notes

- Key mechanical correction to the earlier validated facts: arming the network buffer requires an **active browsing context** — the working sequence is open `about:blank` → arm `list_network_requests` → navigate.
- CI cannot exercise the live path (no macOS Safari session); the `--from-file` mode plus the captured-body fixtures pattern is the intended CI story, listed in the findings doc's ADR questions.


davidwkeith:
## Code review

Reviewed the diff (`scripts/import/browser/{wix-network,safari-mcp,page-functions}.mjs` + the findings doc). This is explicitly a prototype not wired into the import skill, but flagging real correctness issues for whenever it's promoted:

**1. Race condition: fixed 5s `settleMs` silently degrades real blog posts to "not a blog page"** — `wix-network.mjs` `extractPostViaNetwork()`. After `navigate_to_url` the code waits a hardcoded 5000ms then lists network requests once. The findings doc itself documents a delivery generation that's "adapter XHR only" with no warmup-data fallback. If that XHR lands after 5s (slow/throttled network), `adapterReq` isn't found, the warmup-data path also finds nothing, and `buildPost` returns `null` — indistinguishable from a genuine static page. The caller then silently falls back to the DOM path per the documented contract, quietly losing fidelity with no retry or signal that anything went wrong.

**2. List items silently drop non-paragraph content** — the `ORDERED_LIST`/`BULLETED_LIST` case only collects children with `type === 'PARAGRAPH'` into `itemText`, and only descends into `_LIST$`-typed children. An `IMAGE`, `TABLE`, `BLOCKQUOTE`, or `BUTTON` node inside a list item (plausible in a numbered how-to post — exactly the kind of content this module targets) is neither rendered nor recursed into; it just vanishes. This contradicts the stated philosophy in the adjacent `default:` case ("Unknown container nodes: descend so nothing silently disappears").

**3. Unguarded `.find()` on MCP response shape** — `adapterList.requests.find(...)` and `listed.requests.find(...)` assume `.requests` always exists, with no optional chaining or try/catch (unlike the adjacent adapter-JSON parsing, which is wrapped). If `list_network_requests` ever returns an error-shaped payload (e.g. the doc's own noted "No active browsing context" arming failure), this throws an uncaught `TypeError` instead of degrading to the documented null/DOM-fallback contract.

**4. `textToMarkdown` doesn't escape markdown metacharacters** — comment says "Escape nothing." Literal `[`, `]`, `` ` ``, or `*` in real prose (e.g. "see note [1]") pass through unescaped and can be misinterpreted as links/code/emphasis by downstream markdown renderers.

**5. Doc/code mismatch** — the findings doc's fidelity table claims network mode captures "semantic `BOLD`/`LINK`/`ANCHOR` decorations," but `textToMarkdown` only handles `BOLD`, `ITALIC`, and `LINK` — no `ANCHOR` handling exists. If Wix Ricos uses `ANCHOR` for in-page/TOC links, they'd be silently dropped, contradicting the doc's own claim.

Minor: the `setTimeout` in `safari-mcp.mjs#request()` isn't cleared on a normal resolve/reject — harmless (guarded no-op when it fires), just unnecessary garbage for a script issuing many quick requests.

Nothing else stood out — the arming dance, adapter-vs-warmup fallback structure, and Ricos→Markdown conversion for tables/headings/collapsibles look solid for what's explicitly scoped as an exploratory prototype.


---

scripts/import/browser/page-functions.mjs
```diff
@@ -0,0 +1,292 @@
...
+    // Detect block boundaries (different parent block element)
+    const blockParent = node.parentElement?.closest('div, p, h1, h2, h3, h4, h5, h6, li, blockquote');
+    if (blockParent !== lastParent && currentBlock.length > 0) {
+      blocks.push(currentBlock.join(' '));
+      currentBlock = [];
+    }
+    lastParent = blockParent;
+
+    // Check if this is a heading
+    const heading = node.parentElement?.closest('h1, h2, h3, h4, h5, h6, [role="heading"]');
```
davidwkeith: **Headings with more than one text node get fragmented, and links inside headings lose their `href`.**

The heading branch here fires per text node and immediately pushes a new `#`-line, rather than accumulating the heading's inline content the way body paragraphs do:

```js
const heading = node.parentElement?.closest('h1, h2, h3, h4, h5, h6, [role="heading"]');
if (heading) {
  if (currentBlock.length > 0) { blocks.push(currentBlock.join(' ')); currentBlock = []; }
  const level = heading.tagName?.match(/H(\d)/)?.[1] || '2';
  blocks.push('#'.repeat(Number(level)) + ' ' + text);
  lastParent = blockParent;
  continue;
}
```

For `<h1>Hello <strong>World</strong></h1>`, the TreeWalker yields two text nodes ("Hello ", parent `h1`; "World", parent `strong` whose `closest('h1,...')` also matches). Each one independently satisfies `if (heading)` and gets pushed as its own line, producing:

```
# Hello
# World
```

instead of the intended `# Hello World`. The same early `continue` also means the link-wrapping logic just below (`if (link) currentBlock.push(`[${text}](${link.href})`)`) never runs for text inside a heading, so `<h2><a href="...">Linked Heading</a></h2>` loses its href entirely and becomes plain heading text.

Any heading with inline formatting (bold/italic/em/a) will hit this — not an edge case for real blog content. Suggest accumulating heading text into a buffer keyed by the heading element (similar to `currentBlock`) and flushing once when `blockParent` changes, the same way paragraph blocks are handled.

---

scripts/import/browser/safari-mcp.mjs
davidwkeith, line 1: This file is a byte-for-byte copy of the one reviewed in #397 (as the PR description itself notes), and it carries the same two unresolved correctness issues flagged there:

1. **No `'error'` listener on `this.child.stdin`** — only spawn-level failures are caught via `this.child.on('error', ...)`; an EPIPE from writing to a stdin whose reader is gone (e.g. the Safari window was closed) throws an unhandled stream error and crashes the whole import process instead of surfacing a `SafariMcpError`. (See https://github.com/Anglesite/anglesite/pull/397#discussion_r3522538863.)
2. **No `'exit'`/`'close'` handler on `this.child`** — if safaridriver dies while requests are pending, nothing rejects the in-flight promises; each just waits out its own timeout (up to 60s), turning a crash into a long silent hang during a `--site`/multi-URL run instead of a fast error. (See https://github.com/Anglesite/anglesite/pull/397#discussion_r3522538930.)

Since the PR description flags that this branch and #397's need reconciling at merge, worth fixing it once wherever it lands first so the fix doesn't have to be re-applied to both copies.

---

scripts/import/browser/wix-network.mjs
davidwkeith, line 1: `has('BOLD') && decorations.find((d) => d.type === 'BOLD')?.fontWeightValue >= 700` looks suspicious: it only renders `**bold**` when the `BOLD` decoration object also carries a numeric `fontWeightValue >= 700`. If Wix's Ricos `BOLD` decoration is typically just `{ type: 'BOLD' }` with no `fontWeightValue` field (which is the more common shape for a boolean-style decoration), this condition is always false and bold formatting silently never renders — contradicting the findings doc's fidelity table, which claims semantic `BOLD`/`LINK`/`ANCHOR` decorations are recovered.

Worth double-checking against one of the captured payloads (e.g. via `--from-file`) that `fontWeightValue` is actually present on `BOLD` decorations in practice, since this determines whether the bold-recovery claim in the findings doc holds up.

---

davidwkeith (commented):
Reviewed the network-extraction prototype. The Ricos→Markdown converter and adapter/warmup source selection look sound overall (tables, lists, collapsibles, images all handled sensibly), and the exploratory scope/GO-NO-GO framing is clear. Left three comments: a real fragmentation bug in the shared `extractContentSrc` heading handling (drops links, splits multi-node headings), a note that `safari-mcp.mjs` is a literal copy of #397's file and inherits its two already-flagged unhandled-failure issues, and a question about whether the `fontWeightValue >= 700` gate on BOLD decorations matches the real Ricos payload shape.

---

davidwkeith:
Addressed all three review comments in 9bfe167.

**1. Heading fragmentation (`scripts/import/browser/page-functions.mjs`)**
Rewrote the heading branch to accumulate inline text into `currentBlock` (same as paragraphs), tracking the active heading element (`headingEl`) and only flushing the `#`-prefixed line once the walker moves past it (on the existing block-boundary check). The link-wrapping logic (`node.parentElement?.closest('a[href]')`) now runs unconditionally before the heading check, so links inside headings keep their href.

Added `test/wix-page-functions.test.js` (jsdom) with a regression test for `<h1>Hello <strong>World</strong></h1>` → `"# Hello World"` (previously produced `"# Hello\n\n# World"`), a link-in-heading case, and a check that heading text doesn't bleed into the following paragraph. Confirmed these 3 cases fail against the pre-fix code and pass against the fix.

**2. safari-mcp.mjs stdin/exit handling**
Checked — PR #397 only *adds* this file (same bugs, unfixed), so there was no landed fix to mirror; implemented independently, consistent with the file's existing style:
- Added `this.child.stdin.on('error', ...)` so an EPIPE from a closed Safari window rejects in-flight promises instead of crashing the process.
- Added `this.child.on('exit', ...)` that rejects all pending requests immediately (via a shared `failPending` helper also used by the existing `'error'` handler) instead of letting each wait out its own up-to-60s timeout.

**3. BOLD decoration check (`scripts/import/browser/wix-network.mjs`)**
No captured Ricos payload fixtures exist in-repo (checked `test/fixtures/`, the findings doc, and PR #397's fixture) to verify `fontWeightValue` presence directly. Per the public Ricos rich-content schema, `BOLD` is a boolean-style decoration (`{ type: 'BOLD' }`) — `fontWeightValue` isn't part of that decoration's documented shape, so the `>= 700` gate meant bold markdown never fired, silently contradicting the findings doc's claim that BOLD is recovered. Changed the condition to `has('BOLD')` alone, matching the simpler common shape and making the findings-doc claim actually true.

**Tests:** Full suite green — 127 files / 2823 tests, including the new heading-fragmentation test. Also reran `npm run build:agent-skills` (clean, only the expected `jsdom` addition to the bundled `agent-skills/update/references/package.json`) since I added `jsdom` as a devDependency to write a real DOM-driven regression test instead of a source-text-only assertion (the pattern used elsewhere in this repo before jsdom existed as a dependency).

PR #401, opened 2026-07-04
feat(import): Safari MCP rendered-extraction backend for Wix + Squarespace

davidwkeith:
## Summary

Adds Apple's Safari MCP server (`safaridriver --mcp`, Safari Technology Preview 247+) as the **preferred rendered-page extraction backend** for website imports, with Playwright as the unchanged cross-platform fallback — and gives Squarespace imports their **first rendered tier** (design tokens, accordion content, gallery images).

- **`scripts/import/browser/safari-mcp.mjs`** — ~150-line stdio JSON-RPC client that spawns `safaridriver --mcp` directly as a child process. No MCP client configuration, no `.mcp.json`: the backend is a library subprocess, exactly like Playwright. Distinguishes not-installed / not-enabled / session-failed / page-failure / timeout; handles unexpected child death (owner quits the visible Safari window) by failing pending calls fast.
- **`scripts/import/browser/safari-driver.mjs`** — extraction CLI with the same `{tokens, content}` contract as `wix-playwright.mjs`. Batches many URLs in one session (NDJSON out, per-URL error lines), `--check` with exit codes 0/2/3/4 for skill branching, WebKit-native `get_page_content` rescue when the injected extractor comes back empty.
- **`scripts/import/browser/page-functions.mjs`** — the browser-context extractors moved out of the Wix driver and shared by both backends (one source of truth), extended with Squarespace selectors.
- **Import skill Step 1a.2** — resolves `RENDER_BACKEND` (safari → playwright → none); owner-facing copy explains the visible Safari window and the one-time "Allow remote automation" toggle.
- **Docs** — Squarespace `?format=json` endpoint (validated live; structured page JSON via plain curl) + rendered-tier guide; Wix Safari section; ADR-0024; manual live-test protocol in docs/dev/testing.md.

Also fixes a **pre-existing bug**: `extractContentSrc` threw `ReferenceError: fullPage is not defined` on every call (duplicated block testing an undefined variable), so the Playwright content path silently always fell back to curl+regex. First test in the repo that actually executes the extractor in a DOM.

## Why

Wix and Squarespace are the lowest-fidelity import sources. Safari removes the ~150 MB per-project chromium download on macOS, presents a real-Safari fingerprint to bot-throttling hosts, and unlocks Squarespace design-token extraction (previously none). Validated live: cami-demo.squarespace.com → `--color-bg: #c8a47e`, Poppins, full body; wix.com/blog → complete content with 17 full-resolution `static.wixstatic.com` URLs.

## Test plan

- CI: full suite (2838 tests) including new fake-safaridriver tests that exercise real spawn/stdio JSON-RPC, all exit codes, error classification, child-death mid-batch, and the empty-body rescue — no Safari needed in CI.
- Live (macOS, STP 247, automation enabled): the four-step protocol in docs/dev/testing.md, run twice during development, all green.
- Per-task spec+quality reviews plus a whole-branch review; all findings fixed and re-verified.


scripts/import/browser/safari-mcp.mjs
```diff
@@ -0,0 +1,145 @@
...
+      const timer = setTimeout(() => {
+        if (this.pending.has(id)) {
+          this.pending.delete(id);
+          reject(new SafariMcpError('timeout', `${method} timed out after ${timeoutMs}ms`));
+        }
+      }, timeoutMs);
+      timer.unref?.();
+      this.pending.set(id, { resolve, reject });
+    });
+    this.child.stdin.write(JSON.stringify({ jsonrpc: '2.0', id, method, params }) + '\n');
```
davidwkeith, line 122: **No 'error' listener on `child.stdin` — a write after the child dies can crash the whole process.**

`#request()` guards against writing to an already-closed session (`this.closed || !this.child`, line 99), but there's an unavoidable async gap: `safaridriver` can exit and break the stdio pipe *before* Node delivers the `'close'` event (which is what sets `this.closed`/`this.child = null`, lines 67-69). Any `#request()` call made in that window still sees `closed === false` and proceeds to `this.child.stdin.write(...)` (here, and again at line 93-95 for the `notifications/initialized` message). Writing to a broken pipe emits an `'error'` event on the stream; with no listener attached, Node's default behavior is to throw and crash the process instead of surfacing a clean `SafariMcpError('session-failed')` for that one page.

Suggest attaching `this.child.stdin.on('error', () => {})` (or routing it through the same rejection path as the `'close'`/`'error'` handlers on the child itself) so a broken pipe degrades to a per-page error instead of taking down the whole import run.

---

scripts/import/browser/safari-mcp.mjs
```diff
@@ -0,0 +1,145 @@
...
+    const args = ['--mcp'];
+    this.child = this.binaryPath.endsWith('.mjs')
+      ? spawn(process.execPath, [this.binaryPath, ...args], { stdio: ['pipe', 'pipe', 'ignore'] })
+      : spawn(this.binaryPath, args, { stdio: ['pipe', 'pipe', 'ignore'] });
+    this.child.on('error', (err) => {
+      for (const { reject } of this.pending.values()) {
+        reject(new SafariMcpError('session-failed', err.message));
+      }
+      this.pending.clear();
+    });
```
davidwkeith: The `child.on('error', ...)` handler rejects in-flight `pending` calls, but — unlike the `'close'` handler right below it (lines 67-75) — it never sets `this.closed = true` or `this.child = null`. Node doesn't guarantee `'close'` fires synchronously (or at all, in every case) right after `'error'`, so there's a window where the session is effectively dead but the "is this session dead?" guard in `#request()` (line 99: `this.closed || !this.child`) still says no, letting a subsequent call proceed against a broken child.

Consider a single `markDead(reason)` helper called from both handlers so "session is dead" becomes one state transition instead of two independently-maintained listeners.

---

scripts/import/browser/safari-driver.mjs
```diff
@@ -0,0 +1,146 @@
...
+      const wantStyles = !contentOnly && (stylesOnly || index === 0);
+      const tokens = wantStyles ? await extractStyles(mcp) : null;
+      const content = stylesOnly ? null : await extractContent(mcp, fullPage);
+
+      console.log(JSON.stringify({ url, tokens, content }));
+      successes++;
+    } catch (err) {
+      if (err instanceof SafariMcpError && err.code === 'not-enabled') {
+        console.error(err.message);
+        process.exit(EXIT.NOT_ENABLED);
```
davidwkeith, line 167: When `not-enabled` is hit mid-batch, this exits the process immediately without printing an NDJSON line for the URLs after the current one in `urls` — they never appear in the output at all (not even as an `{"url","error"}` line). SKILL.md's documented contract ("a line with `error` falls back per-page like a Playwright timeout", Step 1a.2) has no chance to apply to those pages, since no line is ever emitted for them; the orchestrating agent has no signal that they were skipped rather than, say, silently succeeding or never being part of the batch.

Since `not-enabled` really does mean "stop, nothing further will work," consider still emitting `{"url": u, "error": "..."}` for each of the remaining not-yet-processed URLs before exiting, so downstream consumers can distinguish "attempted and failed" from "never attempted."

---

scripts/import/browser/safari-driver.mjs
```diff
@@ -0,0 +1,146 @@
...
+      await mcp.call('wait_for_navigation', {}, 30000).catch(() => {});
+
+      const expanded = Number(
+        await mcp.call('evaluate_javascript', { expression: iife(expandAccordionsSrc) }),
+      );
+      if (expanded > 0) await sleep(500);
+
+      // Design tokens come from the homepage only (first URL), mirroring the
+      // Playwright driver's "extract styles from the homepage" rule.
+      const wantStyles = !contentOnly && (stylesOnly || index === 0);
```
davidwkeith: Design tokens are only ever attempted on `index === 0` (the homepage, per SKILL.md's "put the homepage URL first" instruction). If that specific page throws — a redirect loop, a timeout, anything caught by the `catch` at line 134 — the loop moves on to the rest of the batch, but `tokens` stays `null` for every subsequent URL too, even though `wantStyles` is never re-evaluated against a later successful page. A batch of 50 pages where only the homepage fails to load ends up with zero design tokens despite 49 successful renders.

Worth considering a fallback: if index 0 fails, attempt styles on the next successful page instead of giving up on tokens for the whole run.

---

scripts/import/browser/safari-mcp.mjs
```diff
@@ -0,0 +1,145 @@
...
+   * setting is off, ('page-failure') for other tool errors.
+   */
+  async call(name, args = {}, timeoutMs = 60000) {
+    const result = await this.#request('tools/call', { name, arguments: args }, timeoutMs);
+    const text = (result.content || [])
+      .filter((c) => c.type === 'text')
+      .map((c) => c.text)
+      .join('\n');
+    if (result.isError) {
+      if (/allow remote automation/i.test(text) && /WebDriverErrorDomain|Could not create a session/i.test(text)) {
```
davidwkeith, line 138: The `not-enabled` classification (which SKILL.md's Step 1a.2 maps to exit code 3 and a dedicated one-time "turn on Allow remote automation" prompt) hinges entirely on two substrings both appearing in Apple's raw, human-readable WebDriver error text. ADR-0024 itself flags that the MCP tool set/error text "may change without notice" across Safari Technology Preview builds. If a future build rewords this (or the OS is running in a non-English locale), the regex silently stops matching, `call()` throws generic `page-failure` instead, `--check` falls into the `SESSION_FAILED` (4) branch, and per SKILL.md line ~260 that's treated as "Safari unavailable, fall through to Playwright silently" — so a fixable one-time permission issue on the owner's own Mac gets misdiagnosed as "Safari isn't there," and the owner never sees the prompt that would fix it.

No CI coverage can catch this since the fake fixture was authored to match this exact regex. Consider whether `safaridriver --mcp` exposes anything more structured (an error code / `data.type`) to key off instead of parsed text.

---

scripts/import/browser/safari-driver.mjs
```diff
@@ -0,0 +1,146 @@
...
+    // Rescue: WebKit-native extraction. Defaults truncate paragraphs to 15
+    // words and strip URL params — override both.
+    const rescueRaw = await mcp.call('get_page_content', {
+      format: 'markdown',
+      region: 'entire_page',
+      maxWordsPerParagraph: 5000,
+      shortenURLs: false,
+    });
+    const rescued = JSON.parse(rescueRaw);
+    const body = rescued.content || '';
```
davidwkeith: The empty-body rescue reads `rescued.content` from the WebKit-native `get_page_content` tool's response — a shape owned by Apple, not this repo. If a future Safari/STP build names this field differently (e.g. `text` instead of `content`), `body` silently becomes `''` and the "rescue" quietly reproduces the exact empty-body problem it exists to fix, with no exception thrown. This is the same failure mode as the pre-existing `extractContentSrc` `ReferenceError` bug this PR fixes elsewhere — a silent fallback that produces empty output instead of an error.

Since only the fake fixture (authored to match this exact assumption) exercises this path in CI, it'd be worth at least asserting `typeof rescued.content === 'string'` and logging/erroring loudly if the shape is unexpected, rather than defaulting to `''`.

---

scripts/import/browser/safari-mcp.mjs
```diff
@@ -0,0 +1,145 @@
...
+    if (envPath.endsWith('.mjs')) return envPath; // test fixture — trusted
+    return supportsMcp(envPath) ? envPath : null;
+  }
+  for (const path of CANDIDATE_PATHS) {
+    if (supportsMcp(path)) return path;
+  }
+  return null;
+}
+
+export class SafariMcp {
```
davidwkeith, line 46: `SafariMcp` hand-rolls a full stdio JSON-RPC MCP client from scratch (id-correlated `pending` map, readline framing, `initialize` handshake, `notifications/initialized`, `tools/call`). The repo already depends on `@modelcontextprotocol/sdk` (`^1.28.0`, in `dependencies`, not just dev) and uses it server-side in `server/index-tools.mjs` (`McpServer`, `StdioServerTransport`) — the same package also ships client-side primitives (`Client`, `StdioClientTransport` in `client/stdio.js`) that implement exactly this lifecycle: spawning a child process, the initialize handshake, request/response correlation, and timeouts.

Reimplementing this by hand means protocol edge cases the SDK already handles (id/type edge cases, cancellation, batching) have to be independently discovered and fixed here rather than inherited from upstream. Worth a look at whether `Client` + `StdioClientTransport` can replace this file's hand-rolled framing — CLAUDE.md's own "MCP-first" preference (currently phrased around Cloudflare provisioning) argues for using the SDK's tooling over reimplementing protocol plumbing where it's already a dependency.

---

scripts/import/wix/wix-playwright.mjs
```diff
@@ -402,9 +86,10 @@ async function main() {
   const url = args.find((a) => !a.startsWith('--'));
```
davidwkeith, line 86: `--fullPage` is now wired end-to-end in both drivers (this file's CLI here, and `safari-driver.mjs`'s `flags.has('--fullPage')`), and the commit history calls this out as "`--fullPage` parity" hardening. But grepping `skills/import/SKILL.md` and `docs/import/wix.md`, `--fullPage` only ever appears in the flag-list documentation (e.g. `[--content-only|--styles-only|--fullPage]`) — nothing in the actual orchestration (Step 3a's homepage-branding extraction, in particular) instructs the model to actually pass `--fullPage` when extracting the homepage. The pre-existing `ReferenceError` this PR fixes (which only threw when `fullPage` was true) is now fixed, but with no live caller ever setting the flag, the header/footer extraction path it unlocks stays effectively dead in real imports.

Worth either wiring Step 3a to pass `--fullPage` for the homepage pass, or noting in SKILL.md that it's reserved for a specific future use so it's clear this isn't an oversight.

---

davidwkeith (commented):
Automated review focused on correctness in the new Safari MCP client/driver, plus one reuse note. Highest-severity items: an unguarded `stdin` write that can crash the process on a broken pipe, a session-death-tracking gap between the `'error'`/`'close'` handlers, a mid-batch `not-enabled` exit that silently drops NDJSON lines for remaining URLs, and design tokens being hard-pinned to the homepage with no fallback if it fails. Also flagged: the brittle double-regex `not-enabled` classification, an unverified assumption about the `get_page_content` rescue response shape, the hand-rolled JSON-RPC client vs. the already-depended-on `@modelcontextprotocol/sdk` client primitives, and `--fullPage` having no live caller despite being wired end-to-end in both drivers.

---

davidwkeith:
Addressed the concrete correctness bugs from the review:

1. **Unguarded stdin write (safari-mcp.mjs)** — attached `child.stdin.on('error', ...)` so a broken pipe (safaridriver exiting before Node delivers `close`) degrades to a per-call rejection instead of crashing the process on an unhandled stream error.

2. **`error` handler didn't mark the session dead (safari-mcp.mjs)** — introduced a single `#markDead(reason)` private method called by both the `error` and `close` handlers (and now the stdin error handler too), so "session is dead" is one consistent state transition instead of two independently maintained code paths.

3. **Mid-batch `not-enabled` exit dropped remaining URLs (safari-driver.mjs)** — before exiting on `not-enabled`, we now emit an `{"url","error"}` line for every URL not yet attempted, so NDJSON consumers get one line per input URL as documented, instead of the tail of the batch silently vanishing. Added a fixture mode (`not-enabled-second-url`) + test covering this.

4. **Design tokens only ever attempted on page 0, no fallback (safari-driver.mjs)** — style extraction is now isolated from content extraction: if the homepage's style pass throws, content extraction for that page still proceeds, and later pages keep retrying token capture until one succeeds, tracked via a `tokensCaptured` flag. Added a fixture mode (`FAKE_TOKENS_FAIL_URL`) + test.

5. **Unvalidated `get_page_content` rescue shape (safari-driver.mjs)** — now asserts `typeof rescued.content === 'string'` and throws a descriptive error if WebKit's response shape doesn't match, instead of silently defaulting to `''` and reproducing the empty-body bug this path exists to fix.

6. **`--fullPage` has no real caller** — added a short doc note in `skills/import/SKILL.md` (and regenerated the `agent-skills/` export) clarifying it's reserved for a future full-page extraction pass and its absence from Step 3a is intentional.

Deliberately deferred (acknowledged, not implemented in this pass):

- **Replacing the hand-rolled stdio JSON-RPC client with `@modelcontextprotocol/sdk`'s `Client`/`StdioClientTransport`** — real improvement, but a large refactor of the whole transport layer; better done as its own focused PR with its own test pass rather than folded into a bug-fix pass.
- **Regex-based `not-enabled` classification on Apple's raw error text** — ADR-0024 already flags this as unstable across STP builds; there's no structured error code to key off instead until Apple exposes one, so there's no better fix available right now, just a known limitation to keep watching.

All 2840 tests pass (`npm test`), including 5 new/updated cases in `test/safari-driver.test.js` covering the mid-batch NDJSON gap fix and the token-fallback fix.

PR #402, opened 2026-07-04
test(optimize-images): raise timeout for scaffold e2e tests to 120s

davidwkeith:
## Summary
- The two tests in `ai-optimize in a scaffolded site (#320)` (`test/optimize-images-packaging.test.js`) shell out to `scripts/scaffold.sh` and take ~9.5-9.8s in isolation — right against Vitest's default 10s timeout.
- Under full-suite parallelism or a loaded machine, they intermittently time out (observed failing at 60s wall in a full-suite run on 2026-07-03, passing at 9.83s in isolation).
- Gave both tests (`scaffolds server/optimize-images.mjs alongside the CLI script` and `runs the optimize script against a fixture JPEG without ERR_MODULE_NOT_FOUND`) an explicit 120s timeout as the third `it()` argument.

## Test plan
- [x] `npx vitest run test/optimize-images-packaging.test.js` — 5/5 passed
- [x] `npm test` — full suite, 2819/2819 passed

### DIFF
--- a/test/optimize-images-packaging.test.js
+++ b/test/optimize-images-packaging.test.js
@@ -83,8 +83,8 @@ describe.skipIf(!hasZsh)('ai-optimize in a scaffolded site (#320)', () => {
   });
 
   // Scaffolding + running the CLI takes ~10s alone; give these room to breathe
-  // under full-suite parallelism and loaded CI runners.
-  const SCAFFOLD_TIMEOUT = 120_000;
+  // under full-suite parallelism and loaded CI runners (~60s observed worst case).
+  const SCAFFOLD_TIMEOUT = 60_000;
 
   it('scaffolds server/optimize-images.mjs alongside the CLI script', () => {
     execFileSync('/bin/zsh', [SCAFFOLD_SCRIPT, '--yes', tmpDir], { stdio: 'pipe' });
--- a/test/optimize-images-packaging.test.js
+++ b/test/optimize-images-packaging.test.js
@@ -82,11 +82,15 @@ describe.skipIf(!hasZsh)('ai-optimize in a scaffolded site (#320)', () => {
     rmSync(tmpDir, { recursive: true, force: true });
   });
 
+  // Scaffolding + running the CLI takes ~10s alone; give these room to breathe
+  // under full-suite parallelism and loaded CI runners.
+  const SCAFFOLD_TIMEOUT = 120_000;
+
   it('scaffolds server/optimize-images.mjs alongside the CLI script', () => {
     execFileSync('/bin/zsh', [SCAFFOLD_SCRIPT, '--yes', tmpDir], { stdio: 'pipe' });
     expect(existsSync(join(tmpDir, 'server', 'optimize-images.mjs'))).toBe(true);
     expect(existsSync(join(tmpDir, 'scripts', 'optimize-images.ts'))).toBe(true);
-  });
+  }, SCAFFOLD_TIMEOUT);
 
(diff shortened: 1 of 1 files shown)


test/optimize-images-packaging.test.js
```diff
@@ -82,11 +82,15 @@ describe.skipIf(!hasZsh)('ai-optimize in a scaffolded site (#320)', () => {
     rmSync(tmpDir, { recursive: true, force: true });
   });
 
+  // Scaffolding + running the CLI takes ~10s alone; give these room to breathe
+  // under full-suite parallelism and loaded CI runners.
+  const SCAFFOLD_TIMEOUT = 120_000;
```
davidwkeith: Nit: 120s is a big blast radius for a test that normally finishes in ~10s — if the scaffold step ever genuinely hangs (rather than just being slow under contention), each of the two tests now eats up to 2 minutes before failing (4 min total) instead of failing fast. The PR description's own data (9.8s isolated, 60s observed under full-suite load) suggests something in the 30–60s range would already give plenty of headroom while still failing reasonably quickly on a real hang. Not a blocker, just worth a sanity check on the margin.

---

test/optimize-images-packaging.test.js
davidwkeith, line 1: Coordination note (not an issue in this diff itself): PR #400, also open, targets flakiness in this exact same `it()` block by adding `--no-alt` to the `execFileSync(tsx, ...)` call further down in this test — root-causing the timeout to the on-device `fm` alt-text cold-load pass (10–20s) rather than to full-suite CI contention. The two fixes don't textually conflict, and this timeout bump would incidentally also cover #400's scenario, but if both land it's worth deciding whether you want one fix or both, and whether 120s should come back down once/if `--no-alt` also lands — otherwise a generous timeout ends up masking a slow path that #400 diagnoses more precisely.

---

davidwkeith (commented):
Small, focused test-timeout fix — the diff itself is correct and the reasoning (sync `execFileSync` calls that can genuinely take longer under full-suite parallelism/loaded CI, not truly hang) checks out. CI is green (all 4 checks passing). Two non-blocking notes left inline:

1. 120s is a generous margin vs. the observed 60s worst case — worth double-checking that's intentional rather than just "big enough to be safe," since a real hang would now cost up to 4 minutes across the two tests.
2. PR #400 (also open) fixes a related flakiness cause in the same test via `--no-alt`; worth reconciling which fix(es) you actually want landed so the timeout doesn't end up masking what #400 diagnoses more precisely.

No changes required to merge as-is.

---

davidwkeith:
Lowered the scaffold e2e timeout from 120s to 60s per review feedback (`SCAFFOLD_TIMEOUT` in `test/optimize-images-packaging.test.js`, shared by both affected tests). 60s matches the documented observed worst case under full-suite CI load with essentially zero margin above it, while still halving the worst-case failure-detection latency versus 120s.

Ran the two affected tests plus the full suite locally after this change — all 5 tests in `test/optimize-images-packaging.test.js` pass (the fixture-JPEG test completed in ~2-3s locally), and the full suite (126 files / 2819 tests) passes clean.

Coordination note: sibling PR #400 fixes flakiness in this same test at its root cause, by adding `--no-alt` to the `execFileSync(tsx, ...)` call to skip the on-device `fm` alt-text cold-load pass (10-20s), which is the likely actual source of the timeout pressure rather than general CI contention. These two fixes don't conflict — this timeout bump would incidentally also cover #400's scenario. If both land, it may be worth revisiting whether the timeout can be lowered further once #400's root-cause fix is in, but that's a call to make once both are merged and reconciled together rather than guessing at it now.

PR #406, opened 2026-07-05
fix(seo-audit): detect dist/client for Workers Static Assets builds

### LINKED ISSUE #390: seo-audit.ts hardcodes distDir='dist', false-positive missing sitemap/robots on Workers Static Assets builds
## What happened

`npm run ai-seo` (scripts/seo-audit.ts) reports "No sitemap.xml / sitemap-index.xml found in dist/" and "No robots.txt found in dist/" even though both files exist and are correctly generated.

## Root cause

The Astro Cloudflare Workers (Static Assets) adapter outputs the client build to `dist/client/`, not directly to `dist/`. `scripts/seo-audit.ts` hardcodes `distDir: "dist"` (both the CLI default at the bottom of the file and the `runAudit` options default), so `findSitemapFile()` and the robots.txt check look in the wrong directory and always report missing files on Workers-adapter projects.

Interestingly, the page-title/word-count checks in the same script still work, because `walkHtml(distDir)` walks recursively and finds the HTML under `dist/client/`, but the resulting page paths incorrectly include a leading `client/` segment (e.g. reported as `/client/about/` instead of `/about/`), because the relative path is computed against `dist` instead of `dist/client`.

## Steps to reproduce

1. Scaffold an Anglesite site using the Cloudflare Workers (Static Assets) output target
2. Run `npm run build`
3. Run `npm run ai-seo`
4. Observe false "missing sitemap.xml" and "missing robots.txt" warnings, and page paths prefixed with `client/`

## Expected behavior

The script should detect the actual client output directory (e.g. check for `dist/client` and use it if present, or accept a `--distDir` flag) so sitemap/robots checks and page-path reporting are accurate for Workers Static Assets projects.

## Environment

- Anglesite version: 1.0.2
- Node: v22.14.0
- OS: Darwin (macOS)

davidwkeith:
## Summary
- `seo-audit.ts` hardcoded `distDir: "dist"`, but the `@astrojs/cloudflare` adapter (Workers Static Assets) writes the client build to `dist/client/`. This caused false-positive "missing sitemap.xml" / "missing robots.txt" warnings, and page paths reported with a leading `client/` segment (e.g. `/client/about/` instead of `/about/`).
- Added `resolveDistDir()`, which auto-detects a `client/` subdirectory under the given base dir and audits it instead when present.
- Added a `--distDir` CLI flag so the base build directory can be overridden explicitly.
- Missing-sitemap/missing-robots messages now report the actual resolved directory instead of a hardcoded `dist/`.

Fixes #390

## Test plan
- [x] `npx vitest run tests/seo-audit.test.ts` — added coverage for `resolveDistDir` and for `runAudit` finding sitemap/robots/pages under a `dist/client/` layout via real temp directories; all 20 tests pass.
- [x] Full suite: 2877/2878 passing (the one pre-existing failure is an unrelated environment issue — missing `tsx`/`jsdom` install for `optimize-images-packaging.test.js`, untouched by this change).

### DIFF
--- a/tests/seo-audit.test.ts
+++ b/tests/seo-audit.test.ts
@@ -1,4 +1,7 @@
-import { describe, it, expect } from "vitest";
+import { mkdtempSync, mkdirSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { describe, it, expect, afterEach } from "vitest";
 import {
   htmlPathToUrlPath,
   isNoindex,
@@ -6,6 +9,8 @@ import {
   extractDescription,
   hasJsonLd,
   exitCodeFor,
+  resolveDistDir,
+  runAudit,
   type SeoAuditReport,
 } from "../template/scripts/seo-audit.js";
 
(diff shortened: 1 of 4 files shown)


tests/seo-audit.test.ts
```diff
@@ -79,6 +84,64 @@ describe("hasJsonLd", () => {
...
+      "<html><head><title>Home</title><meta name=\"description\" content=\"Home page\"></head></html>",
+    );
+    writeFileSync(join(clientDir, "sitemap-index.xml"), "<urlset></urlset>");
+    writeFileSync(join(clientDir, "robots.txt"), "User-agent: *\nAllow: /\n");
+
+    const report = runAudit({ distDir: tmpDir, siteUrl: "https://example.com" });
+
+    expect(report.issues.some((i) => i.code === "missing-sitemap")).toBe(false);
+    expect(report.issues.some((i) => i.code === "missing-robots")).toBe(false);
+    expect(report.pagesAudited).toBe(1);
```
davidwkeith, line 128: Nice coverage for the sitemap/robots resolution, but this doesn't assert the other bug mentioned in the PR description — page paths previously reported with a leading `client/` segment (e.g. `/client/about/` instead of `/about/`). Since `pageData`/`urlPath` isn't exposed on `SeoAuditReport`, consider adding a case that drives `htmlPathToUrlPath` (or asserts via a duplicate-title/description issue whose `page` field would leak the prefix) so a regression on that half of the fix would actually fail a test.

---

davidwkeith (commented):
The `seo-audit.ts` fix itself is solid — `resolveDistDir` is simple, well-tested, and the CLI/`RunAuditOptions` plumbing is correct. One thing worth fixing before/along with this, though:

**`pre-deploy-check.ts` has the identical bug, but in the security-critical path, and this PR doesn't fix it.**

`template/scripts/pre-deploy-check.ts` computes `distDir = resolve(siteDir, "dist")` (in `runScan()`, line ~372) and `const DIST = "dist"` (CLI entry, line ~558), then walks those directly with its own local `walkHtml`/`walkAll` for the four *mandatory* scans: PII (line ~401), tokens (line ~443), Keystatic admin routes (line ~469), and the CLI-path duplicates at lines ~586/609/636. None of these go through `resolveDistDir`.

Per this PR's own summary and `astro.config.ts` (`adapter: isDev ? undefined : cloudflare(...)`), every production build uses the Cloudflare adapter and lands actual output under `dist/client/`, not `dist/`. That means in a normal deploy, `walkHtml(distDir)`/`walkAll(distDir)` here finds **zero files**, and the PII/token/third-party-script/Keystatic scans silently pass with nothing to check — even if the real output under `dist/client/` contains exposed emails, phone numbers, API keys, or unauthorized scripts. This is exactly the class of check `CLAUDE.md` calls non-negotiable ("four mandatory scans before deploying to main").

Notably, the SEO audit call in the same file (`runSeoAudit({ distDir, ... })`, imported from `seo-audit.ts`) *is* fixed transitively by this PR, since `runAudit` now resolves the client subdir internally — but the scans that don't route through `runAudit` are not. Suggest exporting `resolveDistDir` from `seo-audit.ts` and applying it to `distDir`/`DIST` in both `runScan()` and the CLI section of `pre-deploy-check.ts` before the `walkHtml`/`walkAll` calls, as a follow-up (or folded into this PR, since it's the same root cause).

Left one inline nit on the new test file as well.

---

davidwkeith:
Both points addressed in c939f89:

- **Security-critical scans**: exported `resolveDistDir` from `seo-audit.ts` and applied it to `distDir` in `runScan()` and `DIST` in the CLI entry point of `pre-deploy-check.ts`, before all `walkHtml`/`walkAll` calls (PII, tokens, third-party scripts, Keystatic routes). Added a regression test in `tests/pre-deploy-scan.test.ts` that plants PII under `dist/client/` and confirms `runScan` catches it.
- **Test nit**: added a case in `tests/seo-audit.test.ts` that drives a nested `dist/client/about/index.html` through `runAudit` and asserts the resulting `missing-title` issue's `page` field is `/about/`, not `/client/about/` — so a regression on the page-path half of the fix would fail a test.

PR #407, opened 2026-07-05
feat(server): add pagination/filtering to list_content and list_annotations

### LINKED ISSUE #392: list_content and list_annotations MCP tools return unbounded results — add pagination/field selection
## Summary

Two MCP tools return their full result set as a single unbounded blob, unlike `apply_edit`, which correctly windows its before/after diff to a 200-char pad instead of returning full file content (`server/apply-edit-dispatcher.mjs:55-66`):

- **`list_content`** (`server/index-tools.mjs:50-212`) returns the entire site inventory — `{pages, posts, images}` — in one response, 1–10 KB depending on site size, with no filtering.
- **`list_annotations`** returns the full array of unresolved annotations with no cap.

An agent asking a narrow question ("what blog posts exist?") currently pays for the whole site inventory — images, pages, and posts together — every time.

## Proposed fix

Add optional parameters to `list_content`:
- `type`: filter to `pages` | `posts` | `images`
- `limit` / `offset`: paginate large result sets
- `fields`: return only the requested fields per entry (e.g. title + path, not full metadata)

Same shape for `list_annotations` (cap + optional filter by resolved state).

## Why this matters

Tool results are the dominant token cost in agentic sessions — unlike a one-time system-prompt cost, every unnecessary byte in a tool result is paid on every turn that includes it in context. `apply_edit`'s windowed-diff design is the right pattern to follow here.

davidwkeith:
## Summary

Closes #392. `list_content` and `list_annotations` previously returned their entire result set as a single unbounded blob on every call — an agent asking "what blog posts exist?" paid for the whole site inventory (pages + posts + images) every time. Tool results are the dominant token cost in agentic sessions, so this follows the windowed-response pattern `apply_edit` already uses for its before/after diff.

- **`list_content`** (`server/list-content.mjs`, `server/index-tools.mjs`): new optional params
  - `type`: `"pages" | "posts" | "images"` — scan and return only that bucket (skips walking the other two entirely, not just trimming the response)
  - `limit` / `offset`: paginate each returned bucket
  - `fields`: project each entry down to a whitelist of keys (e.g. `["title", "filePath"]`)
- **`list_annotations`** (`server/annotations.mjs`, `server/index-tools.mjs`): new optional params
  - `resolved`: include resolved annotations too (default stays unresolved-only)
  - `limit` / `offset`: paginate

All new params are optional and additive — omitting them reproduces the exact current behavior, so this is not a breaking change for existing callers (including the Anglesite-app's `ContentListing.parse`).

## Test plan

- [x] New unit tests in `test/list-content.test.js` (type filter, per-bucket limit/offset, out-of-range offset, fields projection, unknown-field handling)
- [x] New unit tests in `tests/annotations.test.ts` (resolved filter, limit/offset, combined with existing path filter)
- [x] `npx vitest run test/list-content.test.js tests/annotations.test.ts tests/mcp-server.test.ts test/mcp-http-transport.test.js` — 54/54 passed
- [x] Full `npm test` — 2885/2886 passed (the one failure, `optimize-images-packaging.test.js`, is a pre-existing local environment issue — missing `tsx`/`jsdom` binaries — unrelated to this change)

### DIFF
--- a/server/index-tools.mjs
+++ b/server/index-tools.mjs
@@ -75,9 +75,15 @@ export function buildServer(projectRoot) {
     "List unresolved feedback annotations",
     {
       path: z.string().optional().describe("Filter by page path"),
+      resolved: z
+        .boolean()
+        .optional()
+        .describe("Include resolved annotations too. Default false (unresolved only)."),
+      limit: z.number().int().positive().optional().describe("Max annotations to return"),
+      offset: z.number().int().nonnegative().optional().describe("Skip this many before applying limit"),
     },
-    ({ path }) => {
-      const annotations = listAnnotations(projectRoot, path);
+    ({ path, resolved, limit, offset }) => {
+      const annotations = listAnnotations(projectRoot, path, { resolved, limit, offset });
       return { content: [{ type: "text", text: JSON.stringify(annotations) }] };
     },
   );
@@ -139,9 +145,20 @@ export function buildServer(projectRoot) {
   server.tool(
     "list_content",
     "List the site's pages, article-like content-collection entries (posts, notes, episodes, experiments), and images under public/images, as structured JSON. Read-only; the filesystem is the source of truth.",
-    {},
-    () => {
-      const listing = listContent(projectRoot);
+    {
+      type: z
+        .enum(["pages", "posts", "images"])
+        .optional()
+        .describe("Return only this bucket; the other two come back empty. Default: all three."),
+      limit: z.number().int().positive().optional().describe("Max entries to return per bucket"),
+      offset: z.number().int().nonnegative().optional().describe("Skip this many entries per bucket before applying limit"),
+      fields: z
+        .array(z.string())
+        .optional()
+        .describe("Project each entry down to only these field names, e.g. [\"title\", \"filePath\"]"),
+    },
+    ({ type, limit, offset, fields }) => {
+      const listing = listContent(projectRoot, { type, limit, offset, fields });
       return { content: [{ type: "text", text: JSON.stringify(listing) }] };
     },
   );
--- a/server/list-content.mjs
+++ b/server/list-content.mjs
@@ -29,16 +29,38 @@ const IMAGE_EXTENSIONS = new Set([".jpg", ".jpeg", ".png", ".webp", ".gif", ".sv
  */
 const ARTICLE_COLLECTIONS = new Set(["posts", "notes", "episodes", "experiments"]);
 
+const BUCKET_SCANNERS = { pages: scanPages, posts: scanPosts, images: scanImages };
+
 /**
  * @param {string} projectRoot absolute path to the site root
+ * @param {object} [options]
+ * @param {"pages"|"posts"|"images"} [options.type] scan only this bucket; the other two come
+ *   back empty instead of being walked (#392) — saves the filesystem walk, not just the tokens.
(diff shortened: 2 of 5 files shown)


server/index-tools.mjs
```diff
@@ -75,9 +75,15 @@ export function buildServer(projectRoot) {
     "List unresolved feedback annotations",
```
davidwkeith, line 75: Nit: this description is now stale — it still says "List unresolved feedback annotations" even though `resolved: true` can now return resolved ones too. Worth updating to something like "List feedback annotations (unresolved by default)" so the tool's top-level description matches the new `resolved` param.

---

server/list-content.mjs
```diff
@@ -29,16 +29,38 @@ const IMAGE_EXTENSIONS = new Set([".jpg", ".jpeg", ".png", ".webp", ".gif", ".sv
...
+    result[bucket] = entries;
+  }
+  return result;
+}
+
+/** Project `obj` down to only the requested keys. Unknown keys are silently ignored. */
+function pick(obj, keys) {
+  const out = {};
+  for (const key of keys) {
+    if (key in obj) out[key] = obj[key];
```
davidwkeith, line 61: `key in obj` walks the prototype chain, so a `fields` list containing something like `"constructor"` or `"toString"` would pass this check and copy an inherited function onto `out` (harmless in practice since `JSON.stringify` drops function values, but worth using `Object.hasOwn(obj, key)` — or `Object.prototype.hasOwnProperty.call(obj, key)` — instead of `in` to keep this an explicit own-property projection rather than relying on the JSON-serialization side effect to make it safe).

---

server/index-tools.mjs
```diff
@@ -75,9 +75,15 @@ export function buildServer(projectRoot) {
...
     {
       path: z.string().optional().describe("Filter by page path"),
+      resolved: z
+        .boolean()
+        .optional()
+        .describe("Include resolved annotations too. Default false (unresolved only)."),
+      limit: z.number().int().positive().optional().describe("Max annotations to return"),
```
davidwkeith, line 82: `.positive()` rejects `limit: 0` (same for `list_content`'s `limit` below). That's a reasonable default, but if an agent ever wants "just tell me whether anything matches" (count-only / existence check) without pulling rows, `limit: 0` is a natural way to ask for that and it'll currently fail schema validation instead of returning an empty array. Minor — `.nonnegative()` would allow it without changing any existing behavior, since `limit: 1+` callers are unaffected.

---

davidwkeith (commented):
Reviewed the implementation (`server/list-content.mjs`, `server/annotations.mjs`, `server/index-tools.mjs`) and tests. The pagination/filtering logic itself is correct: `type` genuinely skips scanning the other buckets rather than just trimming the response, `offset`/`limit` slice in the right order, the `resolved` filter composes correctly with the existing `path` filter, and omitting all new params reproduces the old behavior exactly (verified against the "no options" tests). No blocking issues — left three minor/nit comments inline (a stale tool description, a `key in obj` vs. own-property nit, and a `limit: 0` edge case). One more general note: the new params are covered at the `listContent`/`listAnnotations` unit level but not through an actual MCP `server.tool` call with `type`/`fields`/`resolved` args — the existing `mcp-server.test.ts` cases only exercise the no-args path, so the zod schema wiring for the new options isn't exercised end-to-end. Not blocking, just a coverage gap worth closing.

---

PR #409, opened 2026-07-05
fix(skills): split oversized skill bodies for progressive disclosure

### LINKED ISSUE #393: Audit skills for oversized bodies; split heavy reference material out for progressive disclosure
## Summary

The 56 skill directories under `skills/` total ~666 KB of skill-body content (loaded in full whenever a skill is invoked), against ~5.6 KB of always-loaded frontmatter descriptions (loaded once per session, avg 99 bytes/skill — this part is well-disciplined already).

The average skill body is ~12 KB. Some are likely much larger — provider comparison tables, DNS record examples, full API payload samples, and similar reference material that's only needed in specific branches of the skill's workflow are common culprits for skill-body bloat, and they inflate the context cost of every invocation regardless of which branch the task actually needs.

## Proposed fix

1. Run `du -a skills/*/SKILL.md | sort -rn | head -20` (or equivalent) to find the actual heaviest bodies — this issue doesn't have that data yet, only the aggregate.
2. For skills found to carry large reference sections that are only needed in some invocations, split them: keep SKILL.md lean (the core workflow + a short pointer), move the reference material into a linked file the agent reads on demand via its own tools.
3. Skills that are already lean (most, going by the frontmatter discipline) don't need changes — this is a targeted audit, not a blanket rewrite.

## Why this matters

Skill bodies are loaded into context on invocation, not just referenced — an oversized body costs tokens on every turn of every session that touches it, even for tasks that only need the first paragraph.

davidwkeith:
## Summary

Fixes #393. The linked issue's comment includes a byte-count audit of all 56 skill bodies and identifies four concrete split candidates — this PR implements exactly those four splits.

- `skills/stats/SKILL.md` (38.2 KB → 32.8 KB) — moved the zone HTTP analytics fallback (Step 2B, only used when Web Analytics RUM is unavailable) to `skills/stats/zone-http-fallback.md`
- `skills/check/SKILL.md` (32.9 KB → 28.6 KB) — moved the "Deployed worker drift" sub-procedure (gated on `CLOUDFLARE_ACCOUNT_ID` being set) to `skills/check/deployed-worker-drift.md`
- `skills/deploy/SKILL.md` (33.5 KB → 30.2 KB) — moved Step 4 domain setup (buy/transfer/point/subdomain — only relevant on first deploy) to `skills/deploy/domain-setup.md`
- `skills/convert/SKILL.md` (33.3 KB → 29.5 KB) — moved Step 3.5 dynamic route page generation (tag/category/author/date-archive Astro templates, only fires when `PAGINATION_PAGES` is non-empty) to `skills/convert/dynamic-route-pages.md`

Each `SKILL.md` keeps a short pointer in place of the inlined block (`Read ${CLAUDE_PLUGIN_ROOT}/skills/<name>/<file>.md and follow the instructions...`), matching the progressive-disclosure pattern the `import` skill already uses for its platform docs. `agent-skills/` was regenerated via `npm run build:agent-skills` to keep the Open Agent Skills export in sync.

`import` itself was intentionally left alone — the issue's audit flagged it as the model to follow, not a violator (its bulk is a genuinely sequential 9-step workflow, not skippable reference material).

## Test plan

- [x] `npm run build:agent-skills` — clean rebuild, no new warnings, `agent-skills/` matches source
- [x] `npm test` — 2891/2892 pass; the one failure (`optimize-images-packaging.test.js`, missing local `tsx` binary) is a pre-existing environment issue unrelated to this change
- [x] Verified extracted section boundaries match the byte counts reported in the issue's audit table


skills/deploy/domain-setup.md
```diff
@@ -0,0 +1,73 @@
...
+
+If they want a custom domain, surface the domain education prompts from `${CLAUDE_PLUGIN_ROOT}/docs/education-prompts.md` section 2 ("Domain Setup"). Check `.site-config` for each `EDUCATION_<KEY>=shown` flag. Share `DOMAIN_VS_WEBSITE`, `DOMAIN_RENEWAL`, `EMAIL_NOT_AUTOMATIC`, and `TLD_AND_SEO` as a natural aside before diving into options — this is the richest single moment for education. Write the flags to `.site-config` after.
+
+Then determine the right path. Ask: "Do you already own a domain, or do you need to buy one?"
+
+### Option A — Buy a new domain
+
+Before searching, read `${CLAUDE_PLUGIN_ROOT}/docs/domain-guide.md` and check the owner's `BUSINESS_TYPE` in `.site-config`. The right TLD depends on who they are — co-ops should consider .coop, nonprofits should consider .org, environmental orgs should consider .eco. Some mission-aligned TLDs aren't available on Cloudflare; if the best TLD for this owner requires an external registrar, help them register there first and then point nameservers to Cloudflare (Option C below). See the domain guide for the full recommendation table.
```
davidwkeith, line 9: This new file still has raw `${CLAUDE_PLUGIN_ROOT}/docs/domain-guide.md` (here) and `${CLAUDE_PLUGIN_ROOT}/docs/education-prompts.md` (line 7) references. The agent-skills transformer doesn't rewrite `${CLAUDE_PLUGIN_ROOT}` refs inside non-root reference files (only in `SKILL.md` itself), so these ship unresolved in the Open Agent Skills export.

Confirmed by rerunning `npm run build:agent-skills` on this branch: it now emits `⚠ NESTED REF: references/skills/deploy/domain-setup.md still contains ${CLAUDE_PLUGIN_ROOT}`, and the previously-bundled `agent-skills/deploy/references/docs/domain-guide.md` is silently dropped from the export because it's no longer referenced directly from `deploy/SKILL.md`. That's a real functional regression for skills.sh/Open-Agent-Skills consumers of `deploy` — this file is affected by the same "Nested references" limitation currently documented as import-only in `docs/dev/agent-skills.md` line 76.

Given this is a known transformer limitation, at minimum `docs/dev/agent-skills.md` should be updated to note it's no longer import-only (see comment there). Ideally the transformer should also walk `${CLAUDE_PLUGIN_ROOT}` refs inside bundled reference files, not just top-level SKILL.md.

---

skills/check/deployed-worker-drift.md
```diff
@@ -0,0 +1,63 @@
+## Deployed worker drift
+
+Local checks confirm the source compiles — they can't see what's actually running on Cloudflare. When an owner reports "the contact form isn't sending email" or "subscribers aren't being added," the most common cause is local edits that were never deployed. This section verifies each Cloudflare Worker the site uses against its deployed counterpart.
+
+Run this only if `CLOUDFLARE_ACCOUNT_ID` is set in `.site-config` and the active account matches it (see "Cloudflare account alignment" above). If accounts don't match, skip this section — calling Workers APIs against the wrong account would produce misleading results. If `CLOUDFLARE_ACCOUNT_ID` is unset, skip silently.
```
davidwkeith: "(see 'Cloudflare account alignment' above)" is a dangling reference — that section only exists in the parent `skills/check/SKILL.md`, not in this standalone file. Low severity since `SKILL.md` re-checks the gating condition before ever pointing here, but if this file is ever read in isolation (e.g. via the agent-skills export) the "above" pointer goes nowhere. Consider rewording to something that doesn't assume the parent doc's context, e.g. "(see the Cloudflare account alignment check in the main check skill)".

---

skills/stats/zone-http-fallback.md
```diff
@@ -0,0 +1,73 @@
+## Step 2B — Fetch via zone HTTP analytics (fallback)
```
davidwkeith: Minor convention nit: this and the other three new reference files (`skills/check/deployed-worker-drift.md`, `skills/deploy/domain-setup.md`, `skills/convert/dynamic-route-pages.md`) all start with `##` as the top heading. The two existing precedents for this pattern — `skills/import/content-discovery.md` and `skills/import/newsletter-migration.md` — both use `#` for their top heading. Worth bumping these to `#` for consistency with the established standalone-reference-doc convention.

Also, each of the four splits keeps the section heading + 1-3 lead sentences in the parent `SKILL.md` before the "Read .../file.md" pointer, and the new file repeats that same heading + lead text verbatim before continuing. Not harmful, but it means the split isn't quite as clean as the PR description implies — worth trimming the duplication so the parent only has the pointer and the gating condition, with the heading/lead prose living in exactly one place.

---

davidwkeith (commented):
Reviewed the four splits against the actual moved content (diffed `origin/main` vs. this branch, read every extracted file in full, reran `npm run build:agent-skills` and the test suite on this commit). Overall the extractions are byte-for-byte faithful — no truncation or duplication of the actual procedural content, and all four pointers correctly preserve their original gating conditions (`CLOUDFLARE_ACCOUNT_ID`, `PAGINATION_PAGES`, the stats source check). Step numbering in the parent files also stays coherent since content was removed mid-section rather than removing whole steps.

Left inline comments on the three concrete issues found. One more that doesn't anchor to a diff line: `docs/dev/agent-skills.md` line 76 currently says the "Nested references" `${CLAUDE_PLUGIN_ROOT}`-bundling limitation "mostly affects the `import` skill" — this PR extends that same defect to `deploy` (see the `domain-setup.md` comment), so that doc line is now stale and should be updated as part of this change, not left for a follow-up.

Not a blocker, but noting for the record: `deploy` and `convert` remain over the spec's 500-line `SKILL.md` recommendation even after this split (per the build's own warnings), so further splitting may still be worth a follow-up if that threshold matters.

---

PR #410, opened 2026-07-06
feat(mcp): get_component_model tool (Component Editor slice 1)

davidwkeith:
Adds a read-only component-model service: @astrojs/compiler template tree + css-tree style rules + heuristic Props extraction, exposed as the get_component_model MCP tool. App-side consumer: Anglesite-app Component Editor (spec 2026-07-05-component-editor-design.md in Anglesite-app). No behavior change to existing tools.

### DIFF
--- a/server/component-model.mjs
+++ b/server/component-model.mjs
@@ -7,6 +7,7 @@ import { createHash } from "node:crypto";
 import { execFileSync } from "node:child_process";
 import { parse } from "@astrojs/compiler";
 import { parse as parseCss, generate, walk } from "css-tree";
+import { parseProps } from "./props-interface.mjs";
 
 export class ComponentModelError extends Error {
   constructor(reason, message) {
@@ -57,13 +58,34 @@ export async function buildComponentModel(projectRoot, relPath) {
   collectElements(ast, "style", styleElements);
   const styles = styleElements.flatMap((el) => extractRules(el));
 
+  const fmNode = topLevel.find((n) => n.type === "frontmatter");
+  const frontmatter = fmNode
+    ? {
+        source: fmNode.value ?? "",
+        span: [fmNode.position?.start?.offset ?? null, fmNode.position?.end?.offset ?? null],
+        props: parseProps(fmNode.value ?? ""),
+      }
+    : null;
+
+  const scriptElements = [];
+  collectElements(ast, "script", scriptElements);
+  const scriptText = scriptElements
+    .map((el) => (el.children ?? []).find((c) => c.type === "text"))
+    .find((t) => t?.value);
+  const clientScript = scriptText
+    ? {
+        source: scriptText.value,
+        span: [scriptText.position?.start?.offset ?? null, scriptText.position?.end?.offset ?? null],
+      }
+    : null;
+
   return {
     version: fileVersion(projectRoot, source),
     path: relPath,
     template,
-    frontmatter: null, // Task 3
+    frontmatter,
     styles,
-    clientScript: null, // Task 3
+    clientScript,
   };
 }
 
--- a/server/component-model.mjs
+++ b/server/component-model.mjs
@@ -4,7 +4,6 @@
 import { readFileSync } from "node:fs";
 import { join, normalize } from "node:path";
 import { createHash } from "node:crypto";
-import { execFileSync } from "node:child_process";
 import { parse } from "@astrojs/compiler";
 import { parse as parseCss, generate, walk } from "css-tree";
 import { parseProps } from "./props-interface.mjs";
@@ -80,7 +79,7 @@ export async function buildComponentModel(projectRoot, relPath) {
     : null;
 
   return {
-    version: fileVersion(projectRoot, source),
+    version: fileVersion(source),
     path: relPath,
     template,
     frontmatter,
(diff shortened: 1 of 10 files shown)


server/component-model.mjs
```diff
@@ -0,0 +1,201 @@
...
+  }
+}
+
+function fileVersion(projectRoot, source) {
+  try {
+    return execFileSync("git", ["rev-parse", "HEAD"], { cwd: projectRoot, encoding: "utf-8" }).trim();
+  } catch {
+    return "sha256:" + createHash("sha256").update(source).digest("hex").slice(0, 12);
+  }
+}
```
davidwkeith, line 200: `fileVersion` uses the repo's `HEAD` SHA, not the file's actual content state, so it doesn't do what its name (and the `source`-hashing fallback) implies.

Concretely, `recordEdit` in `server/edit-history.mjs` commits every `apply_edit` patch onto the hidden `refs/heads/anglesite/edits` branch *without touching HEAD* (that's the documented design — "without touching the user's HEAD, current branch, index, or working tree"). So after an edit lands on disk via `apply_edit`, `git rev-parse HEAD` in `projectRoot` returns the exact same SHA as before the edit. A caller that fetches `get_component_model` before and after an edit gets an identical `version` even though the file content changed — the opposite of what a version/ETag field is for.

It also ignores the `source` parameter entirely in the primary branch, only using it in the `catch` fallback — that asymmetry is itself a signal something's off here.

Suggest hashing the actual file content unconditionally (`"sha256:" + createHash(...).update(source)...`), or if git is deliberately preferred when available, use something that tracks *this file's* blob (e.g. `git hash-object <path>` against the working tree, or include the working-tree mtime/size) rather than the repo-wide `HEAD` commit.

---

server/props-interface.mjs
```diff
@@ -0,0 +1,24 @@
...
+  const block = frontmatterSource.match(/(?:interface\s+Props|type\s+Props\s*=)\s*\{([\s\S]*?)\n\}/);
+  if (!block) return [];
+  const props = [];
+  for (const line of block[1].split("\n")) {
+    const m = line.match(/^\s*(\w+)(\?)?\s*:\s*([^;]+);?\s*(?:\/\/.*)?$/);
+    if (m) props.push({ name: m[1], type: m[3].trim(), optional: Boolean(m[2]), default: null });
+  }
+  const destructure = frontmatterSource.match(/const\s*\{([\s\S]*?)\}\s*=\s*Astro\.props/);
+  if (destructure) {
+    for (const part of destructure[1].split(",")) {
```
davidwkeith, line 16: Splitting the destructure block on `,` breaks any default value that itself contains a comma — e.g. `const { items = [1, 2, 3] } = Astro.props` splits into `"items = [1"`, `" 2"`, `" 3]"`. The first chunk matches the regex and sets `items.default = "[1"` (a truncated, syntactically-wrong value) instead of leaving it `null` or capturing the full literal.

This is worse than the documented "anything fancier yields []" fallback for the `interface Props` block above — there the failure mode is empty/no-knobs, but here it's silently wrong data returned as if it were valid, which the Component Editor would show to the user as the real default.

Worth at least guarding against unbalanced brackets/parens when splitting (skip setting `default` if the chunk has more `[`/`{`/`(` than closing ones), or documenting this as a known gap like the block above does.

---

server/index-tools.mjs
```diff
@@ -140,6 +141,35 @@ export function buildServer(projectRoot) {
...
+    "Parse an .astro component into a structured, read-only model: template node tree with source spans, frontmatter Props interface, scoped style rules, and client script zone. Used by the app's Component Editor.",
+    {
+      path: z.string().describe("Component path relative to the project root, e.g. src/components/Card.astro"),
+    },
+    async ({ path }) => {
+      try {
+        const model = await buildComponentModel(projectRoot, path);
+        return { content: [{ type: "text", text: JSON.stringify(model) }] };
+      } catch (err) {
+        const reason = err instanceof ComponentModelError ? err.reason : "parse-failed";
```
davidwkeith: Any non-`ComponentModelError` thrown here (e.g. a `TypeError` from a bug in `extractRules`/`NodeBuilder`, or css-tree throwing something unexpected) gets reported to the app as `reason: "parse-failed"`, i.e. "this component's Astro/CSS syntax is invalid." That's misleading for what's actually an internal bug, and would push the app toward "can't edit this file" UX instead of surfacing a real defect. Consider a distinct `reason` (e.g. `"internal-error"`) for the unexpected-error branch so parse failures and tool bugs aren't conflated.

---

davidwkeith (commented):
Reviewed `get_component_model` (Component Editor slice 1). The parsing/tree-building logic is solid and well-tested, but the `version` field is functionally broken given how `apply_edit` already works in this codebase — see the inline comment on `fileVersion`. Two smaller correctness/robustness notes on the heuristic Props parser and error-reason labeling.

---

server/component-model.mjs
davidwkeith: Fixed in f035009 — `fileVersion` now hashes file content unconditionally (`sha256:` prefix, git call removed entirely). Verified the premise: `recordEdit` commits onto `refs/heads/anglesite/edits` without moving HEAD, so the old repo-SHA version was constant across every apply_edit. Added a test that the version changes when file content changes and is stable when it doesn't.

server/props-interface.mjs
davidwkeith: Fixed in f035009 — the destructure chunk now must pass a balanced-brackets-and-quotes check before its value is recorded; otherwise `default` stays `null` (unknown) rather than truncated-but-plausible data. Covered by a new test: `items = ["a", "b"]` and `label = "hello, world"` yield `null`, while `count = 2` in the same destructure still comes through. Capturing the full literal (bracket-aware splitting) is left for the fast-follow list.

server/index-tools.mjs
davidwkeith: Fixed in f035009 — unexpected exceptions now report `reason: "internal-error"`; only `ComponentModelError` reasons (`invalid-input`/`read-failed`/`parse-failed`) pass through. The Swift client treats `reason` opaquely (it surfaces the whole error payload as `toolFailed`), so no consumer change needed. Branch is a one-line ternary; no direct test since forcing a non-ComponentModelError through the real server requires mocking the module behind a spawned process.

PR #412, opened 2026-07-06
fix(mcp): component-model parser gaps (slice 1 fast-follows)

### LINKED ISSUE #411: component-model parser gaps (slice 1 fast-follows)
Triaged from the Component Editor slice 1 final review (#410):

- **`<Fragment>` dropped with its entire subtree** — `toNode`'s default case discards fragment nodes, so the outline silently omits their content. Map fragments to a pass-through node (or hoist children).
- **`is:global` / non-CSS `<style>` (scss) produce garbage rules** — css-tree error-recovery yields rows like selector `\"$x: 1; .a\"`. Detect non-plain-CSS style blocks (`lang`/preprocessor attrs) and skip them instead.
- **Comma-containing prop defaults: capture instead of null** — the balanced-brackets guard (post-review f035009) correctly avoids truncated values, but a bracket-aware split could return the real literal (`[\"a\", \"b\"]`).
- **Fixture test for nested `<style>`/`<script>` zone filtering** — behavior is wider than tested (zones filtered at any depth); pin it before slice 2 relies on zone extraction.

davidwkeith:
## Summary

Closes #411 — the four parser gaps triaged from the Component Editor slice 1 final review (#410):

- **`<Fragment>` pass-through** — `toNode` now maps fragment nodes to a `kind: "fragment"` template node (attrs preserved, children mapped) instead of dropping the entire subtree in the default case.
- **Non-CSS `<style>` blocks skipped** — `extractRules` returns no rules when the style element has a `lang` attribute other than `css` (scss/less), so css-tree error-recovery can't emit garbage rows like selector `"$x: 1; .a"`. `is:global` blocks with plain CSS still extract (pinned by test).
- **Comma-containing prop defaults captured** — the naive comma split + balanced-brackets null guard is replaced with a bracket/quote-aware top-level splitter, so defaults like `["a", "b"]`, `"hello, world"`, and `{ x: 1, y: 2 }` come through intact instead of `null`.
- **Nested zone filtering pinned** — fixture test asserting `<style>`/`<script>` are filtered from the template tree at any depth while their zones (styles, clientScript) are still extracted.

## Test plan

- `npx vitest run tests/component-model.test.ts` — 14 tests pass (5 new/updated, all watched fail first)
- Full suite: only pre-existing failures unrelated to this change (`tests/build-agent-skills.test.ts`, `test/optimize-images-packaging.test.js` — both fail identically at HEAD)

### DIFF
--- a/server/component-model.mjs
+++ b/server/component-model.mjs
@@ -94,6 +94,8 @@ function collectElements(node, name, out) {
 }
 
 function extractRules(styleElement) {
+  const lang = (styleElement.attributes ?? []).find((a) => a.name === "lang")?.value;
+  if (lang && lang.toLowerCase() !== "css") return []; // scss/less etc.: css-tree error-recovery emits garbage rows
   const textChild = (styleElement.children ?? []).find((c) => c.type === "text");
   if (!textChild?.value) return [];
   const baseOffset = textChild.position?.start?.offset ?? 0;
@@ -157,6 +159,8 @@ class NodeBuilder {
       case "component":
       case "custom-element":
         return this.#make(n, "component", n.name);
+      case "fragment":
+        return this.#make(n, "fragment", null);
       case "expression":
         return { ...this.#base(n), kind: "expression", tag: null, attrs: [], children: [] };
       case "text": {
@@ -165,7 +169,7 @@ class NodeBuilder {
         return { ...this.#base(n), kind: "text", tag: null, attrs: [], text: value.slice(0, 80), children: [] };
       }
       default:
-        return null; // comment, doctype, fragment wrappers
+        return null; // comment, doctype
     }
   }
   #make(n, kind, tag) {
--- a/server/component-model.mjs
+++ b/server/component-model.mjs
@@ -95,7 +95,7 @@ function collectElements(node, name, out) {
 
 function extractRules(styleElement) {
   const lang = (styleElement.attributes ?? []).find((a) => a.name === "lang")?.value;
-  if (lang && lang.toLowerCase() !== "css") return []; // scss/less etc.: css-tree error-recovery emits garbage rows
+  if (lang && lang.trim().toLowerCase() !== "css") return []; // scss/less etc.: css-tree error-recovery emits garbage rows
   const textChild = (styleElement.children ?? []).find((c) => c.type === "text");
   if (!textChild?.value) return [];
   const baseOffset = textChild.position?.start?.offset ?? 0;
--- a/server/props-interface.mjs
+++ b/server/props-interface.mjs
@@ -13,27 +13,36 @@ export function parseProps(frontmatterSource) {
   }
   const destructure = frontmatterSource.match(/const\s*\{([\s\S]*?)\}\s*=\s*Astro\.props/);
   if (destructure) {
-    for (const part of destructure[1].split(",")) {
+    for (const part of splitTopLevel(destructure[1])) {
       const dm = part.match(/^\s*(\w+)\s*=\s*(.+?)\s*$/s);
       if (!dm) continue;
       const prop = props.find((p) => p.name === dm[1]);
-      if (prop && isBalanced(dm[2].trim())) prop.default = dm[2].trim();
+      if (prop) prop.default = dm[2].trim();
     }
   }
   return props;
 }
 
-// A comma inside a default (array/object/call/string literal) makes the naive
-// comma-split above produce a truncated chunk. Rather than return that
-// silently-wrong value, require balanced brackets and quotes and leave
-// `default` null (unknown) when the chunk fails the check.
-function isBalanced(value) {
-  const pairs = { "(": ")", "[": "]", "{": "}" };
-  const open = [];
-  for (const ch of value) {
-    if (pairs[ch]) open.push(pairs[ch]);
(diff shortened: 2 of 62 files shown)


server/props-interface.mjs
```diff
@@ -13,27 +13,36 @@ export function parseProps(frontmatterSource) {
   }
   const destructure = frontmatterSource.match(/const\s*\{([\s\S]*?)\}\s*=\s*Astro\.props/);
   if (destructure) {
-    for (const part of destructure[1].split(",")) {
+    for (const part of splitTopLevel(destructure[1])) {
       const dm = part.match(/^\s*(\w+)\s*=\s*(.+?)\s*$/s);
       if (!dm) continue;
       const prop = props.find((p) => p.name === dm[1]);
-      if (prop && isBalanced(dm[2].trim())) prop.default = dm[2].trim();
+      if (prop) prop.default = dm[2].trim();
```
davidwkeith: Removing `isBalanced` drops the safety net it was providing, and it's easy to reproduce a regression: feed `parseProps` a destructure with an unbalanced bracket (very plausible mid-edit in the live Component Editor, e.g. an in-progress array literal), and this now silently assigns garbage to `default` instead of leaving it `null`.

Reproduced against this branch:
```js
const src = `
interface Props {
  items: string[];
  label?: string;
}
const { items = ["a", "b", label = "fallback" } = Astro.props;
`;
parseProps(src);
// => [{ name: "items", ..., default: "[\"a\", \"b\", label = \"fallback\"" }, { name: "label", ..., default: null }]
```

On `main` (pre-PR), the same input yields `items.default === null` and `label.default === "\"fallback\""` — worse in one way (misses `label`) but never silently wrong.

The module's own doc comment says the design goal is "Anything fancier yields [] … never an error" — i.e. prefer `null`/missing over wrong data. `splitTopLevel` is a real improvement for well-formed source, but for malformed/mid-edit source (exactly what a live editor will see) it should fall back to leaving `default: null` rather than assigning whatever fragment the naive split produces. Consider re-adding a balance check (e.g. bail out per-part, or for the whole destructure, when `splitTopLevel` never returns to depth 0) before trusting `dm[2]`.

---

server/component-model.mjs
```diff
@@ -94,6 +94,8 @@ function collectElements(node, name, out) {
 }
 
 function extractRules(styleElement) {
+  const lang = (styleElement.attributes ?? []).find((a) => a.name === "lang")?.value;
+  if (lang && lang.toLowerCase() !== "css") return []; // scss/less etc.: css-tree error-recovery emits garbage rows
```
davidwkeith: Nit, non-blocking: `lang` isn't trimmed before comparing, so `<style lang=" css ">` (stray whitespace, plausible from a hand-edited or pasted attribute) would be treated as non-CSS and its rules dropped even though it's plain CSS. Cheap fix: `lang.trim().toLowerCase() !== "css"`.

---

davidwkeith (commented):
Reviewed the three fixes for #411. The Fragment pass-through, nested style/script zone filtering, and the new `splitTopLevel` bracket/quote-aware comma splitter are all solid and covered by good tests (verified locally — `npx vitest run tests/component-model.test.ts` passes 14/14 on this branch, and I confirmed `<Fragment>`/`<>` both parse to `type: "fragment"` via `@astrojs/compiler` so the new switch case is exercised correctly).

One real issue (inline): dropping `isBalanced` entirely removes the guard that kept malformed/mid-edit destructure text from producing a wrong (not just missing) `default` value — see the repro on `server/props-interface.mjs`. Since this parser feeds a live Component Editor that will see in-progress edits, I'd want that safety net back before merging.

One non-blocking nit inline on the `lang` comparison in `component-model.mjs`.

Separately, and not caused by this diff: CI is currently red on this PR (`test (22)`, `test (24)`, `agent-skills-export`), but it looks like pre-existing drift on `main` — `.claude-plugin/plugin.json` is already at `1.3.0` while the committed `agent-skills/` export still embeds `version: "1.2.0"` in its frontmatter, so `build-agent-skills.test.ts` and the `agent-skills-export` job fail regardless of this PR's changes. Worth a quick `npm run build:agent-skills` + commit on `main` (or as part of whichever branch merges first) so this PR (and any other open PR) can go green.

---

server/props-interface.mjs
davidwkeith: Re-added in e5d7d8b, applied per-part after `splitTopLevel`. The new `isBalanced` is quote/escape-aware (the old one counted raw quote parity, so a legitimate default like `"it's"` would have been rejected). Your repro is now a test case — `items` and `label` both come back `default: null` on the malformed destructure, and the well-formed comma-literal cases still capture intact.

server/component-model.mjs
davidwkeith: Fixed in e5d7d8b — `lang.trim().toLowerCase() !== "css"`, with a test pinning `lang=" css "` as plain CSS.

davidwkeith:
Addressed both review points in e5d7d8b (balance guard re-added, quote-aware; `lang` trimmed — both TDD'd, 16/16 component-model tests pass). Also took your suggestion to fix the CI drift here since this branch is likely to merge first: b198d04 regenerates `agent-skills/` for v1.3.0 (pure version-bump diff, 59 files). `build-agent-skills.test.ts` passes 87/87 locally; CI should go green.

PR #413, opened 2026-07-07
feat: component-style write ops (Component Editor slice 2)

davidwkeith:
## Summary

Part 1 of Component Editor Slice 2 (Anglesite-app#492): the plugin-side write ops for the Styles panel.

- Extracts shared `fileVersion`/`indexCssRules` helpers (span-precise CSS rule indexing) out of `component-model.mjs` so the read model and the new write resolver can't drift.
- Adds four `apply_edit` ops: `set-style-property`, `remove-style-property`, `add-style-rule`, `set-rule-selector` — identified by exact byte span (never by selector text, which `css-tree`'s `generate()` normalizes away from the source).
- Content-hash `baseVersion` staleness check on every write, refused as `"stale"` before any mutation.
- Successful component-style edits piggyback a freshly rebuilt `get_component_model` result on the reply, so the app never needs a second round-trip.
- Full stdio MCP round-trip test proves the real server path, not just unit tests.

Built task-by-task via subagent-driven development (implementer + independent reviewer per task, with an extra-scrutiny pass on the byte-splice resolver — the reviewer independently reproduced compiler offsets to verify a real edge-case bug the implementer caught and fixed in the empty-`<style>` insertion path).

## Test plan

- [x] `npm test` — 138/138 files, 2942/2942 tests passing
- [x] `npm run build:agent-skills` regenerated and its CI-gating test (`tests/build-agent-skills.test.ts`) passes
- [x] Version synced across `package.json`, `.claude-plugin/plugin.json`, `template/package.json`, `CLAUDE.md` per `bin/release.ts`'s convention (bumped by hand here — see note below)
- [ ] Tag/publish `v1.4.0` — **deliberately deferred**, not part of this PR. Needs a go-ahead before `bin/release.ts`'s tag+push step runs.

Closes #492 (plugin half; app-side consumption is a separate Anglesite-app PR).

### DIFF
--- a/server/component-style-edit.mjs
+++ b/server/component-style-edit.mjs
@@ -33,7 +33,7 @@ export async function resolveComponentStyle(projectRoot, edit) {
   try {
     source = readFileSync(absPath, "utf-8");
   } catch (err) {
-    return refuse("no-match", `read ${relPath}: ${err.message}`);
+    return refuse("read-failed", `read ${relPath}: ${err.message}`);
   }
 
   if (fileVersion(source) !== baseVersion) {
@@ -44,7 +44,7 @@ export async function resolveComponentStyle(projectRoot, edit) {
   try {
     ({ ast } = await parse(source, { position: true }));
   } catch (err) {
-    return refuse("invalid-input", `parse ${relPath}: ${err.message}`);
+    return refuse("parse-failed", `parse ${relPath}: ${err.message}`);
   }
 
   const styleElements = [];
--- a/server/component-style-edit.mjs
+++ b/server/component-style-edit.mjs
@@ -1,0 +1,167 @@
+import { readFileSync } from "node:fs";
+import { join, normalize } from "node:path";
+import { parse } from "@astrojs/compiler";
+import { fileVersion } from "./file-version.mjs";
+import { indexCssRules } from "./css-rule-index.mjs";
+
+/**
+ * Write-side resolver for the four Component Editor style ops
+ * (set-style-property, remove-style-property, set-rule-selector,
+ * add-style-rule). Re-indexes the target .astro file's <style> block(s)
+ * fresh from disk via `indexCssRules` (Task 2) and turns the requested op
+ * into a precise byte-range splice — the same `{file, range, replacement}`
+ * shape every other resolver in `patcher.mjs` produces.
+ */
+
+function refuse(reason, detail) {
+  return { refused: true, reason, detail };
+}
+
+export async function resolveComponentStyle(projectRoot, edit) {
+  const { component } = edit;
+  if (!component || typeof component !== "object") {
+    return refuse("invalid-input", "component payload is required for this op");
+  }
+  const { path: relPath, baseVersion, ruleSpan, property, value, selector, media, declarations } = component;
+
+  if (typeof relPath !== "string" || !relPath.endsWith(".astro") || normalize(relPath).startsWith("..") || relPath.startsWith("/")) {
+    return refuse("invalid-input", `not a project-relative .astro path: ${relPath}`);
+  }
+
+  const absPath = join(projectRoot, relPath);
+  let source;
+  try {
+    source = readFileSync(absPath, "utf-8");
+  } catch (err) {
+    return refuse("no-match", `read ${relPath}: ${err.message}`);
+  }
+
+  if (fileVersion(source) !== baseVersion) {
+    return refuse("stale", `${relPath} changed since the model was fetched`);
+  }
+
+  let ast;
+  try {
+    ({ ast } = await parse(source, { position: true }));
+  } catch (err) {
+    return refuse("invalid-input", `parse ${relPath}: ${err.message}`);
+  }
+
+  const styleElements = [];
+  collectStyleElements(ast, styleElements);
+  const rules = styleElements.flatMap((el) => indexCssRules(el));
+
+  switch (edit.op) {
+    case "set-style-property":
+      return applySetStyleProperty(relPath, rules, ruleSpan, property, value);
+    case "remove-style-property":
+      return applyRemoveStyleProperty(relPath, source, rules, ruleSpan, property);
+    case "set-rule-selector":
+      return applySetRuleSelector(relPath, rules, ruleSpan, selector);
+    case "add-style-rule":
+      return applyAddStyleRule(relPath, source, styleElements, selector, media, declarations);
+    default:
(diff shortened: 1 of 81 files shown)


server/component-style-edit.mjs
```diff
@@ -0,0 +1,167 @@
...
+    return refuse("no-match", `read ${relPath}: ${err.message}`);
+  }
+
+  if (fileVersion(source) !== baseVersion) {
+    return refuse("stale", `${relPath} changed since the model was fetched`);
+  }
+
+  let ast;
+  try {
+    ({ ast } = await parse(source, { position: true }));
```
davidwkeith, line 45: **TOCTOU gap between the `baseVersion` check here and the actual write in `apply-edit-dispatcher.mjs`.**

The staleness guard at line 39 validates `fileVersion(source) === baseVersion` against the copy of `source` read a few lines up — but then this function `await`s `parse(source, ...)` (an actual async yield point, since `@astrojs/compiler`'s parser is asynchronous). After this resolver returns `{file, range, replacement}`, `apply-edit-dispatcher.mjs`'s `applyEdit` does a **second, independent** `readFileSync(absPath)` and splices `range` into *that* fresh read — without re-checking the hash.

If a second `apply_edit` call for the same file lands and completes while this one is suspended at the `await parse(...)` point (plausible over the HTTP transport, which can serve concurrent requests, or even two rapid style-panel edits from the app), the dispatcher will splice this call's stale byte offsets into the *other* call's already-written content. That's silent corruption of the source file — exactly the class of bug `baseVersion` was introduced to prevent, just moved one step later in the pipeline.

Every other resolver in `patcher.mjs` is fully synchronous, so this window doesn't exist for them today; it's new here because of the `await parse(...)`. Worth either:
- having the dispatcher re-verify `fileVersion(freshlyReadSource) === baseVersion` immediately before splicing/writing for component-style ops, or
- having `resolveComponentStyle` return the exact `source` it validated so the dispatcher splices that buffer instead of re-reading (and still re-checks before the write wins the race).

---

server/component-style-edit.mjs
```diff
@@ -0,0 +1,167 @@
...
+  if (typeof relPath !== "string" || !relPath.endsWith(".astro") || normalize(relPath).startsWith("..") || relPath.startsWith("/")) {
+    return refuse("invalid-input", `not a project-relative .astro path: ${relPath}`);
+  }
+
+  const absPath = join(projectRoot, relPath);
+  let source;
+  try {
+    source = readFileSync(absPath, "utf-8");
+  } catch (err) {
+    return refuse("no-match", `read ${relPath}: ${err.message}`);
```
davidwkeith: Minor: reusing `"no-match"` for a file read failure (and `"invalid-input"` for a parse failure at line 47) is inconsistent with the sibling `buildComponentModel` in `component-model.mjs`, which uses dedicated `"read-failed"` / `"parse-failed"` reasons for these exact same failure modes on the same file. Since both modules share `fileVersion`/`indexCssRules` specifically so they can't drift, it'd be good for their refusal vocabulary to agree too — otherwise the app can't tell "the rule span you referenced no longer exists" (a real `no-match`) apart from "we couldn't even open/parse the file", and `invalid-input` here would actually mean "the file's own content failed to parse", not "your request payload was malformed" (which is what `invalid-input` means everywhere else in this file).

---

davidwkeith (commented):
Reviewed the plugin-side write ops for the Component Editor's styles panel (`server/component-style-edit.mjs`, `server/css-rule-index.mjs`, `server/file-version.mjs`, plus the dispatcher/schema/patcher wiring).

**Overall:** solid design — extracting `indexCssRules`/`fileVersion` so the read model (`component-model.mjs`) and the new write resolver share the exact same span-indexing logic is the right call, and the test coverage is good (empty `<style>`, no `<style>` at all, media queries, stale/no-match refusals, and a real stdio MCP round-trip). I ran the full suite locally (`npm test`) — all green except two pre-existing failures in `test/apply-edit-dispatcher.test.js` (`write-failed surface`) that also fail on `origin/main` in this sandbox because they're running as root, where chmod-based permission checks don't apply; not a regression from this PR.

Two things worth addressing before merge, left as inline comments:

1. **Race between the `baseVersion` staleness check and the actual disk write** (`component-style-edit.mjs:45`, interacting with `apply-edit-dispatcher.mjs`'s second `readFileSync`). The new async `await parse(...)` step creates a real yield point between validating the hash and performing the splice, and the dispatcher re-reads the file fresh afterward without re-checking it. A concurrent edit landing in that window would get silently spliced with stale byte offsets — undermining the exact protection `baseVersion` was built to provide. Every other resolver in `patcher.mjs` is synchronous, so this gap is new here.

2. **Refusal-reason inconsistency** (`component-style-edit.mjs:36` and the parse-failure branch at line 47): read/parse failures reuse `"no-match"` / `"invalid-input"` instead of the `"read-failed"` / `"parse-failed"` reasons `buildComponentModel` already uses for the identical failure modes on the same file, which muddies what those reasons mean for API consumers.

Neither is a blocker for the design, but #1 in particular seems worth closing before this ships, since it's the specific failure mode the feature is meant to prevent.

Nit (non-blocking): the new-declaration/new-rule text synthesis in `applySetStyleProperty`/`applyAddStyleRule` hardcodes 2-space indentation regardless of the file's actual formatting — matches the project's own style-guide default, so low priority, but could surprise a component authored with different indentation.

---

server/component-style-edit.mjs
davidwkeith: Fixed in 93247f1 — the dispatcher now re-validates fileVersion(source) against component.baseVersion immediately before splicing, using its own fresh readFileSync (which was already happening, just unchecked). This closes the window: a concurrent edit landing during the async parse(...) now gets refused as stale instead of silently splicing stale offsets into the other call's content. Added a test that races a real concurrent write against the async parse gap and asserts both the stale refusal and that the other write survives untouched.

server/component-style-edit.mjs
davidwkeith: Fixed in 93247f1 — read failures now return read-failed and parse failures return parse-failed, matching buildComponentModel's ComponentModelError reasons for the same failure modes. Added a read-failed test; didn't add a dedicated parse-failed test since @astrojs/compiler's parser is lenient enough that I couldn't find a reliably-failing fixture without over-engineering one (no existing precedent for this in component-model.test.ts either) — the reason-code fix itself is still correct and covered by the existing catch block.

PR #415, opened 2026-07-09
fix(mcp): walk JSX embedded in a top-level conditional expression (#414)

### LINKED ISSUE #414: component-model outline extraction fails when component root is a conditional expression
## Summary

The component-model outline/markup extraction returns an empty tree (just a placeholder `{…}` row, no children, no expand action) when an Astro component's entire template body is a single top-level conditional expression instead of a plain element root.

Found while running the slice-1 manual GUI smoke test (Anglesite-app#491) against `Hcard.astro`:

```astro
---
const profile = ...
---
{profile && (
  <footer class="site-identity">
    <div class="h-card">
      ...
    </div>
  </footer>
)}
```

## Repro

1. Open a component whose template is `{someCondition && ( <element>...</element> )}` (no plain top-level element) in the Component Editor.
2. Outline panel shows a single collapsed row with placeholder text `{…}` and no children — `AXDisclosing` is `false` and neither the icon nor the label exposes an expand/press action.
3. Canvas still renders the harness correctly (the runtime render is unaffected), but nothing in the canvas can be selected/mapped back to an outline row since the model has no elements.
4. Source tab works fine and shows the correct raw text — this is scoped to markup/outline extraction, not source loading.

## Isolation

Confirmed via `BaseLayout.astro` in the same site, whose root is `<!doctype html>...<html>...` (plain element, not conditional) — outline renders the full tree correctly, including a nested `<Hcard />` reference. So general outline extraction is fine; the gap is specifically a component whose only top-level node is a logical-AND/conditional expression wrapping a single element.

## Suggested fix direction

`toNode`'s (or equivalent) root-handling in the component-model parser needs to unwrap a top-level `JSXExpressionContainer` guard expression (e.g. `cond && (<el/>)`, and likely the ternary form `cond ? <el/> : null`) down to the JSX element it renders, similar to how #411 handled `<Fragment>` unwrapping — this looks like the same class of gap (parser only walks a plain top-level element) that #411 was triaged from, just a different shape of top-level node.

## Why this matters

Component Editor slice 2 (Anglesite-app#492, Styles panel first write ops) builds directly on outline/zone extraction. Any component using a conditional root (a common pattern for "only render this if data exists") will have an unusable editor until this is fixed.

davidwkeith:
Fixes #414.

## Summary

- `component-model.mjs`'s `NodeBuilder.toNode` hardcoded `children: []` for every `expression` AST node, on the assumption an expression is always a leaf (e.g. `{title}`). That's true for plain interpolation, but an expression's `children` in `@astrojs/compiler`'s AST also carries real embedded JSX — e.g. a conditional root `{profile && (<footer>...</footer>)}` or a mapped list `{items.map(i => <li>{i}</li>)}` — mixed in with "text" pseudo-nodes that are actually raw JS source, not markup.
- Any component whose entire template is a single top-level conditional (a common "only render if data exists" pattern — this is exactly what `Hcard.astro` in the template's microformats set does) got an outline with one childless `{…}` placeholder row and nothing to select or map canvas clicks onto.
- Fix: recurse into an expression's `element`/`component`/`custom-element`/`fragment` children (real embedded JSX) while still dropping its `text` pseudo-children (JS source). Simple expressions like `{title}` are unaffected — their only child is a "text" JS-identifier pseudo-node, which is still filtered out, so they stay leaves.

No Swift-side changes needed — `ComponentOutline.rows(from:)` and `ComponentOutline.node(atLine:column:)` in Anglesite-app already recurse generically over `node.children` regardless of parent kind, so nested JSX now surfaces correctly once the server sends it.

Found via Anglesite-app#491's manual GUI smoke test of the Component Editor against `Hcard.astro`, filed as #414.

## Test plan
- [x] Added a regression test reproducing the exact `Hcard.astro` shape (conditional root wrapping `<footer><div><p>{profile.name}</p></div></footer>`) — walks the full nested tree and confirms the inline `{profile.name}` expression still stays a childless leaf.
- [x] `npx vitest run tests/component-model.test.ts` — 17/17 pass
- [x] `npx vitest run` (full suite) — 138 files / 2945 tests pass

### DIFF
--- a/server/component-model.mjs
+++ b/server/component-model.mjs
@@ -122,7 +122,20 @@ class NodeBuilder {
       case "fragment":
         return this.#make(n, "fragment", null);
       case "expression":
-        return { ...this.#base(n), kind: "expression", tag: null, attrs: [], children: [] };
+        // An expression's `children` mix raw JS source (as "text" pseudo-nodes,
+        // e.g. "profile && (" ) with any JSX actually embedded in it (e.g. a
+        // conditional root `{cond && (<el/>)}` or a mapped list). Only the
+        // latter are real markup — walk those; drop the JS-source text.
+        return {
+          ...this.#base(n),
+          kind: "expression",
+          tag: null,
+          attrs: [],
+          children: (n.children ?? [])
+            .filter((c) => c.type === "element" || c.type === "component" || c.type === "custom-element" || c.type === "fragment")
+            .map((c) => this.toNode(c))
+            .filter(Boolean),
+        };
       case "text": {
         const value = (n.value ?? "").trim();
         if (!value) return null;
--- a/server/component-model.mjs
+++ b/server/component-model.mjs
@@ -107,6 +107,10 @@ function isZoneNode(n) {
   return n.type === "frontmatter" || (n.type === "element" && (n.name === "style" || n.name === "script"));
 }
 
+// AST node types that represent real embedded JSX inside an expression's
+// `children` (as opposed to "text" pseudo-nodes, which are raw JS source).
+const JSX_CHILD_TYPES = new Set(["element", "component", "custom-element", "fragment"]);
+
 class NodeBuilder {
   #next = 0;
   nextId() {
@@ -126,13 +130,18 @@ class NodeBuilder {
         // e.g. "profile && (" ) with any JSX actually embedded in it (e.g. a
         // conditional root `{cond && (<el/>)}` or a mapped list). Only the
         // latter are real markup — walk those; drop the JS-source text.
+        //
+        // A two-branch conditional (`{cond ? <A/> : <B/>}`) surfaces both `<A>`
+        // and `<B>` as siblings here even though only one renders at a time —
+        // acceptable for a static outline/editor tree, but worth knowing if a
+        // future consumer assumes every outline row is on the live page.
         return {
           ...this.#base(n),
           kind: "expression",
           tag: null,
           attrs: [],
           children: (n.children ?? [])
-            .filter((c) => c.type === "element" || c.type === "component" || c.type === "custom-element" || c.type === "fragment")
+            .filter((c) => JSX_CHILD_TYPES.has(c.type))
             .map((c) => this.toNode(c))
             .filter(Boolean),
         };
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,12 @@ All notable changes to this project will be documented in this file.
   content-hash (`baseVersion`) staleness checking. Successful component-style
   edits piggyback a freshly rebuilt `get_component_model` result on the reply.
 
+### Fixed
+- `get_component_model`'s outline extraction no longer drops JSX embedded in
+  a top-level conditional expression (e.g. `{profile && (<footer>...</footer>)}`)
+  — a component whose entire template was one conditional root previously got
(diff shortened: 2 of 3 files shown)


server/component-model.mjs
```diff
@@ -122,7 +122,20 @@ class NodeBuilder {
...
+        // e.g. "profile && (" ) with any JSX actually embedded in it (e.g. a
+        // conditional root `{cond && (<el/>)}` or a mapped list). Only the
+        // latter are real markup — walk those; drop the JS-source text.
+        return {
+          ...this.#base(n),
+          kind: "expression",
+          tag: null,
+          attrs: [],
+          children: (n.children ?? [])
+            .filter((c) => c.type === "element" || c.type === "component" || c.type === "custom-element" || c.type === "fragment")
```
davidwkeith: Nit: this line is quite long (135 cols) and dense compared to the rest of the file. Consider splitting the type check into a small helper (e.g. `isEmbeddedJsx(c)`) or a `Set`/array lookup for readability:

```js
const JSX_CHILD_TYPES = new Set(["element", "component", "custom-element", "fragment"]);
...
children: (n.children ?? [])
  .filter((c) => JSX_CHILD_TYPES.has(c.type))
  .map((c) => this.toNode(c))
  .filter(Boolean),
```

Not blocking.

---

server/component-model.mjs
```diff
@@ -122,7 +122,20 @@ class NodeBuilder {
...
+            .map((c) => this.toNode(c))
+            .filter(Boolean),
+        };
```
davidwkeith, line 147: Good fix for the conditional-root case (`{cond && (...)}`) and it's well covered by the new regression test. One edge case worth being aware of (not necessarily blocking, since it's out of scope for this fix): a two-branch conditional like `{cond ? <A/> : <B/>}` will now surface **both** `<A>` and `<B>` as children of the same `expression` node, even though only one ever renders at a time. That's probably fine for a static outline/editor tree, but if the app's Component Editor assumes each outline row corresponds to something that's actually on the rendered page, it might be worth a follow-up note (or at least a comment here) so it isn't mistaken for a bug later.

---

davidwkeith (commented):
Solid, well-scoped fix. `NodeBuilder.toNode`'s `expression` case previously hardcoded `children: []`, so any component whose template was a single top-level conditional (`{profile && (...)}`) produced an outline with a childless placeholder and nothing to click/map — exactly the `Hcard.astro` bug in #414. The fix correctly distinguishes real embedded JSX (`element`/`component`/`custom-element`/`fragment` children) from the JS-source `text` pseudo-nodes that also live in an expression's `children`, and recurses only into the former. Verified against `#make`'s existing recursive pattern and confirmed no other server module (`selector.mjs`, `patcher.mjs`, `apply-edit-dispatcher.mjs`) special-cases `kind: "expression"` in a way this would break — the Swift-side `ComponentOutline` already walks `node.children` generically, per the PR description.

Left two non-blocking nits inline (a readability suggestion on the filter predicate, and a note about the ternary `{cond ? <A/> : <B/>}` case surfacing both branches as siblings) plus a suggestion to add a `CHANGELOG.md` `### Fixed` entry under `Unreleased` since it patches a bug in the still-unreleased `get_component_model` feature. Test coverage for the reported case looks good (nested element/attrs/leaf-expression assertions all present).

---

PR #416, opened 2026-07-13
feat(mcp): component-structure write ops + Unicode span-corruption fix

davidwkeith:
## Summary
- Adds `apply_edit` component-structure ops (`set-attr`, `remove-node`, `insert-node`) backed by a monotonic-cursor lexical span resolver (`resolveAllSpans`) that never trusts `@astrojs/compiler`'s reported byte offsets — it re-derives every node's true source span from tree shape + a forward lexical walk, refusing (`no-match`) rather than risking silent corruption when a span can't be found.
- Fixes a real `@astrojs/compiler@4.0.0` bug discovered during review: `position.*.offset` is a UTF-8 **byte** offset, not a JS-string (UTF-16) index, so any multi-byte-UTF-8 character (emoji, accented Latin, CJK, etc.) earlier in a component's source silently corrupted the `span`/`loc` of every node after it — in the already-shipped `get_component_model` read model (template nodes, attributes, frontmatter, client script) as well as CSS rule/declaration spans. All of these now derive offsets from the (reliably UTF-16-based) line/column fields instead of trusting `.offset`.

## Test plan
- [x] `npm test` — 142 files / 3004 tests passing
- [x] Unicode regression tests added to `tests/component-node-index.test.ts`, `tests/component-model.test.ts`, and `tests/css-rule-index.test.ts`, confirmed RED before the fix / GREEN after
- [x] `tests/component-structure-edit.test.ts` covers set-attr, remove-node, insert-node, and nested-sibling-collision regressions

### DIFF
--- a/server/apply-edit-schema.mjs
+++ b/server/apply-edit-schema.mjs
@@ -52,6 +52,10 @@ export const editOps = [
   "remove-style-property",
   "add-style-rule",
   "set-rule-selector",
+  "insert-node",
+  "move-node",
+  "remove-node",
+  "set-attr",
 ];
 
 /** The subset of `editOps` that operate on a component's scoped `<style>` via `component`
@@ -64,6 +68,13 @@ export const COMPONENT_STYLE_OPS = new Set([
   "set-rule-selector",
 ]);
 
+/** The subset of `editOps` that operate on a component's template structure (DOM nodes)
+ *  via `component` rather than on a page element via `selector`. */
+export const COMPONENT_STRUCTURE_OPS = new Set(["insert-node", "move-node", "remove-node", "set-attr"]);
+
+/** Union of style and structure component ops — used by the dispatcher's shared checks. */
+export const COMPONENT_OPS = new Set([...COMPONENT_STYLE_OPS, ...COMPONENT_STRUCTURE_OPS]);
+
 /** Structured payload for the four component-style ops. Identifies the target rule by its exact
  *  byte span (from `get_component_model`'s `styles[].span`) plus a `baseVersion` content-hash
  *  guard so a stale client-side model is refused rather than silently mis-patching. */
@@ -75,13 +86,28 @@ export const componentEditSchema = z.object({
     .optional()
     .describe("Identifies an existing rule by its exact byte span from get_component_model's styles[].span. Required for set-style-property, remove-style-property, set-rule-selector. Omitted for add-style-rule."),
   property: z.string().optional().describe("Declaration property name; required for set-style-property and remove-style-property"),
-  value: z.string().optional().describe("Declaration value; required for set-style-property"),
+  value: z.string().nullable().optional().describe("Declaration value; required for set-style-property. For set-attr, a null value means remove the attribute."),
   selector: z.string().optional().describe("New rule's selector for add-style-rule, or the renamed selector for set-rule-selector"),
   media: z.string().nullable().optional().describe("@media condition for add-style-rule; absent/null means no wrapping media query"),
   declarations: z
     .array(z.object({ property: z.string(), value: z.string() }))
     .optional()
     .describe("Initial declarations for add-style-rule"),
+  nodeId: z.string().optional().describe("Identifies an existing template node by its get_component_model id. Required for set-attr, remove-node, move-node's source node."),
+  name: z.string().optional().describe("Attribute name for set-attr"),
+  parentId: z.string().optional().describe("Parent node id for insert-node (the fragment root's id for a top-level insert)"),
+  index: z.number().int().optional().describe("Child index to insert at, for insert-node"),
+  newParentId: z.string().optional().describe("Destination parent node id for move-node"),
+  newIndex: z.number().int().optional().describe("Destination child index for move-node"),
+  node: z
+    .object({
+      kind: z.enum(["element", "component", "slot"]),
+      tag: z.string().optional().describe("HTML tag name (element) or component name (component); omitted for slot"),
+      componentPath: z.string().optional().describe("Project-relative .astro path to import, required when kind=component"),
+      slotName: z.string().optional().describe("Named slot, for kind=slot; omitted means the default slot"),
+    })
+    .optional()
+    .describe("New node spec for insert-node"),
 });
 
 /** The MCP tool's input shape, as passed to `server.tool(name, description, shape, handler)`. */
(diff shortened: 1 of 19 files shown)


server/apply-edit-schema.mjs
```diff
@@ -64,6 +68,13 @@ export const COMPONENT_STYLE_OPS = new Set([
   "set-rule-selector",
 ]);
 
+/** The subset of `editOps` that operate on a component's template structure (DOM nodes)
+ *  via `component` rather than on a page element via `selector`. */
+export const COMPONENT_STRUCTURE_OPS = new Set(["insert-node", "move-node", "remove-node", "set-attr"]);
```
davidwkeith, line 73: **Critical: the new ops are never actually dispatched — `apply_edit` will crash (or return a bogus refusal) for all four.**

`patcher.mjs`'s `resolve()` only special-cases `COMPONENT_STYLE_OPS`:
```js
if (COMPONENT_STYLE_OPS.has(edit.op)) {
  return resolveComponentStyle(projectRoot, edit);
}
```
It never imports `COMPONENT_STRUCTURE_OPS`/`COMPONENT_OPS` or `resolveComponentStructure` from `component-structure-edit.mjs`. `apply-edit-dispatcher.mjs` has the same gap (only imports `COMPONENT_STYLE_OPS`). So a well-formed `insert-node`/`remove-node`/`set-attr`/`move-node` call (no `selector`, since these ops use `component` per the schema) falls through to `resolveMdoc → resolveKeystatic → resolveAstro`, all of which unconditionally do `selector.textContent`.

I verified this empirically against this branch: calling `applyEdit()` with a valid `set-attr` payload on a project with no `.mdoc` files returns a nonsensical `edit-failed`/`no-match`/"no .mdoc files found"; on a project that *does* have a `.mdoc` file (the normal case for any real Anglesite site) it throws an **unhandled `TypeError: Cannot read properties of undefined (reading 'textContent')`** out of `resolveMdoc` (`patcher.mjs:228`), which propagates out of `applyEdit` uncaught.

None of the new tests catch this because they all call `resolveComponentStructure` directly rather than going through `patcher.mjs`'s `resolve()` / `apply-edit-dispatcher.mjs`'s `applyEdit()`. `component-structure-edit.mjs` is fully implemented and unit-tested, but it's dead code from the MCP tool's perspective — `patcher.mjs` needs an early branch (mirroring the `COMPONENT_STYLE_OPS` one) routing `COMPONENT_STRUCTURE_OPS` to `resolveComponentStructure`, plus an integration test that actually calls `applyEdit`/`resolve` for at least one structure op end-to-end.

---

server/component-structure-edit.mjs
```diff
@@ -0,0 +1,715 @@
...
+      return applyMoveNode(relPath, source, byId, rootId, component);
+    default:
+      return refuse("invalid-input", `unsupported component-structure op: ${edit.op}`);
+  }
+}
+
+function applySetAttr(file, source, byId, component) {
+  const { nodeId, name, value } = component;
+  if (typeof nodeId !== "string" || typeof name !== "string") {
+    return refuse("invalid-input", "set-attr requires component.nodeId and component.name");
```
davidwkeith, line 75: `applySetAttr` never checks `node.kind` before treating `node` as tag-shaped. `resolveAllSpans` (used by remove-node/insert-node) explicitly refuses `text` and tagless-`fragment` nodes and documents that `expression` nodes have unreliable spans even via line/column — but `set-attr` skips that discipline entirely and trusts `node.span`/`node.attrs` straight from `byId` for *any* node kind.

Concretely, if `nodeId` refers to:
- a **text** node: `attrs` is always `[]`, so it takes the "insert new attribute" path with `insertAt = node.span[0] + 1 + (node.tag?.length ?? 0)` → `node.tag` is `null` → splices `` name="value"`` one character into the middle of the text content (e.g. `"Hello world"` → `` "H class="x"ello world"``).
- an **expression** node: same insert path, but `node.span` itself can be off-by-one per the compiler quirk documented above `resolveAllSpans` — compounding the corruption.
- a shorthand-**fragment** (`<>...</>`) node: `insertAt = span[0] + 1` lands inside `<>`, producing invalid markup.

None of this is far-fetched — `nodeId` is caller-supplied and `get_component_model` happily hands out ids for text/expression/fragment nodes alongside element/component/slot ones, so a stale or slightly-off client-side selection would silently corrupt the file rather than refuse. Worth adding a `kind` guard here (`element`/`component`/`slot` only, refuse `invalid-input`/`no-match` otherwise) to match the fail-closed posture the rest of this file (correctly) takes. None of the added tests exercise set-attr against a non-element node, so this gap isn't currently caught.

---

davidwkeith (commented):
Reviewed the diff and independently ran the new code paths to check the two inline findings.

**Blocking:** the four new component-structure ops (`insert-node`, `move-node`, `remove-node`, `set-attr`) are added to the `editOps` schema and fully implemented/tested in `component-structure-edit.mjs`, but `patcher.mjs`'s `resolve()` (the function `apply-edit-dispatcher.mjs`'s `applyEdit()` calls) was never updated to route them there — it only special-cases `COMPONENT_STYLE_OPS`. I verified against this branch that calling `apply_edit` with a valid `set-attr` payload on a normal project (one with `.mdoc` content files) throws an unhandled `TypeError` instead of performing the edit; on a project with no `.mdoc` files it returns a nonsensical `no-match`/"no .mdoc files found" refusal. All the new tests call `resolveComponentStructure` directly, so this gap isn't caught by `npm test`. See the inline comment on `apply-edit-schema.mjs` for details — this needs a `COMPONENT_STRUCTURE_OPS` branch in `patcher.mjs` (and an end-to-end test through `applyEdit`) before this is mergeable.

**Worth fixing:** `applySetAttr` doesn't validate the target node's `kind` before treating it as tag-shaped, unlike `resolveAllSpans` (used by insert/remove) which explicitly refuses non-element-shaped nodes. A `nodeId` pointing at a text/expression/fragment node can splice attribute syntax into the middle of running text or into an unreliable expression span. See the inline comment on `component-structure-edit.mjs`.

The Unicode/byte-offset fix itself (the `component-node-index.mjs` line/column rewrite and the lexical `resolveAllSpans` cursor walk) looks solid and well-reasoned — nice test coverage on the nested-sibling-collision and Unicode-precedes-node cases in particular.

---

server/apply-edit-schema.mjs
davidwkeith: Fixed in 05720a1: `patcher.mjs`'s `resolve()` now routes `COMPONENT_STRUCTURE_OPS` to `resolveComponentStructure` (mirroring the `COMPONENT_STYLE_OPS` branch), and `apply-edit-dispatcher.mjs`'s three `COMPONENT_STYLE_OPS` checks (component-payload guard, staleness re-check, model piggyback) now use the `COMPONENT_OPS` union so structure ops get the same treatment as style ops. Added `tests/apply-edit-dispatcher-component-structure.test.ts` covering set-attr end-to-end through `applyEdit` (success + model piggyback, missing-payload refusal, stale refusal, and the concurrent-write race).

server/component-structure-edit.mjs
davidwkeith: Fixed in ebe031f: `applySetAttr` now refuses `invalid-input` unless `node.kind` is `element`/`component`/`slot`, before touching `node.span`/`node.attrs` at all — matching the fail-closed discipline `resolveAllSpans` already applies for insert/remove/move. Added a `kind guard` describe block in `tests/component-structure-edit.test.ts` covering set-attr against a text node, an expression node, and confirming element/component/slot still work.

PR #420, opened 2026-07-17
feat(mcp): add extract-component op (Component Editor slice 5 prep)

davidwkeith:
## Summary

Adds `extract-component` to the `apply_edit` MCP tool: lifts a selected outline subtree out of an existing `.astro` component into a brand-new `src/components/<newName>.astro` file, hoists "obvious" literal props, and replaces the selection in the source file with an instance of the new component plus a matching import. This is plugin-side prep for `Anglesite/Anglesite-app#495` (Component Editor slice 5's "Extract into Component…" affordance) — the design spec's op table already routes it through the existing `apply_edit` tool rather than a new dedicated tool.

This is architecturally novel versus the existing structure/style/frontmatter ops because it's a **two-file write** (every prior op patches exactly one file), which required two explicit design decisions:

### 1. Two-file write ordering + response shape

- **Write order:** the brand-new component file is written **first**, then the source file is patched **second**. If the second write fails, the new file is rolled back with a best-effort `unlink` — cheap and safe because the new file never existed before this op. The reverse order would risk leaving the source file referencing a component that doesn't exist on disk if the second write then failed. There's still no cross-file transaction (this repo's `atomicWrite` only gives single-file atomicity, same accepted limitation as `create-content.mjs`), but this ordering closes the realistic failure window.
- **Git history:** `edit-history.mjs`'s `recordEdit` gains a `files: string[]` mode that stages N files onto **one** `anglesite/edits` commit (via the existing temp-index/`write-tree`/`commit-tree` plumbing, just looped over multiple paths before the single `write-tree` call), so "one semantic op → one commit" holds here too — a single `undo_edit` call reverts the whole extraction. Single-file callers (`{file, range}`) are 100% unchanged.
- **Latent bug fixed as a prerequisite:** `undo-edit.mjs` previously assumed every file in the commit being undone already existed in the parent commit (`git show ${parent}:${file}` unconditionally) — undoing an extract would have crashed on that call for the brand-new file. Fixed to detect (`git cat-file -e`) and delete such files instead of trying to restore content that never existed.
- **Response shape:** `createEditAppliedContent` grows an additive `newFile: string` field (the new file's project-relative path) alongside the existing `file`/`range` (which still describe the SOURCE file's edit). Every other op's reply is unchanged — `newFile` is simply `undefined` and omitted from the JSON for them.
  ```jsonc
  // edit-applied body for extract-component
  {
    "type": "anglesite:edit-applied",
    "id": "...",
    "file": "src/components/Hero.astro",     // the SOURCE file (unchanged field)
    "range": { "start": 0, "end": 123 },      // whole-file replacement, same as remove/insert/move-node
    "commit": "abc123...",                    // ONE commit covering both files
    "model": { /* fresh get_component_model for the SOURCE file */ },
    "newFile": "src/components/CardTitle.astro"  // NEW field, additive only
  }
  ```

### 2. Prop-hoisting heuristic

A literal is "obvious" enough to hoist only when its exact string value occurs **exactly once** among all quoted-attribute-values and text-node contents within the extracted subtree (`kind: "quoted"` per `@astrojs/compiler`'s own attribute-kind enum — `expression`/`shorthand`/`spread`/`template-literal` attrs are never touched; text nodes modeled by `buildTemplateNodeIndex` are by construction always static).

- **Naming:** attribute props are named after the (camelCased) attribute name — `data-size` → `dataSize` — with a reserved-word guard (`class`/`for`/etc. → `classValue`/`forValue`) since a bare destructured binding named `class` is a syntax error. Text props get a generic `text1`, `text2`, … sequence in document order. Collisions (e.g. two different attrs that both camelCase to the same name) get a numeric suffix.
- **Type:** every hoisted prop is `type: "string"` — no number/boolean inference. No prior art for it anywhere in this codebase, and it's not worth the complexity for a first cut.
- **Span discipline:** the extracted subtree's own root boundary is re-derived via `component-structure-edit.mjs`'s `resolveAllSpans` lexical walk — never trusts `node.span` directly, same discipline as `remove-node`/`insert-node`/`move-node`. Individual literal candidates *within* that already-verified boundary use their own `attr.span`/`node.span` directly (matching the already-shipped `set-attr` resolver's precedent), each sanity-checked against its expected text before being trusted — a mismatch just drops that one candidate rather than risking a bad splice.
- **Import hygiene:** reuses `ensureImport`/`parseImports` (`frontmatter-imports.mjs`) for the source file's new import, and separately carries over imports for any component-kind descendants used *inside* the extracted subtree (e.g. extracting a wrapper that itself renders `<Badge ... />`) so the new file doesn't reference an undefined component. Refuses (`invalid-input`) when `newName` would collide with an existing, differently-targeted import in the source file, since `ensureImport` dedups by specifier, not local name. Bare/aliased import specifiers (`"some-lib/Icon.astro"`, `"@components/Button.astro"`) are passed through unchanged rather than mis-rewritten as relative paths.
- **Dynamic content is refused, not copied.** `collectCandidates` already skipped expression-kind nodes and non-`"quoted"` attributes when deciding what to *hoist* — but that alone didn't stop such content from being copied byte-for-byte into the new file whenever it wasn't specifically hoisted. A `{title}` text interpolation, a dynamic attribute binding (`src={imageUrl}`), a spread, or a template literal all reference identifiers that would not be in scope in the new file (which only gets the hoisted-prop destructure), so the new file would silently fail to build. Fixed in a follow-up commit: `findDynamicContent` walks the subtree up front and refuses with `dynamic-expression` (the same reason `patcher.mjs`'s `resolveAstro` already uses for "can't safely act on dynamic content") if it finds any expression-kind node or attribute — matching every sibling structural op's fail-closed convention, rather than attempting to hoist expressions as pass-through props in this same change.

## Other changes

- `component-node-index.mjs`: `attrsOf` now also carries the compiler's own attribute `kind` internally, so the resolver can tell literals from dynamic values. **Not** exposed by `component-model.mjs`'s public `get_component_model` JSON — verified `toPublicNode` whitelists `{name, value}` only, so the wire contract is unchanged.
- `component-structure-edit.mjs`: exports `resolveAllSpans`, `SpanResolutionError`, `escapeAttr`, `importSpecifier`, `VOID_ELEMENTS`, `collectComponentTags` for reuse (export-only diff — no behavior change) rather than a second, drift-prone copy of this tricky lexical-span logic.
- `apply-edit-schema.mjs`: new `extract-component` op string, `COMPONENT_EXTRACT_OPS` Set (added to the `COMPONENT_OPS` union), `newName` field on `componentEditSchema` (reuses the existing `nodeId` field).
- No version bump — per this repo's own process, that happens after review as a separate step.

## Test plan

- [x] `npm test` — full suite passes (150 files / 3082 tests, up from 147/3047 on `main`)
- [x] `tests/component-extract-edit.test.ts` — resolver unit tests: refusals (invalid-input, stale, no-match, already-exists, root-extraction, newName collision, dynamic-expression via text interpolation, dynamic-expression via a dynamic attribute binding), successful extraction with attr+text hoisting, reserved-word renaming, duplicate-value non-hoisting, component-kind extraction with prop hoisting, carried imports (relative, bare/aliased), frontmatter synthesis when the source has none, a sanity check that the dynamic-content refusal doesn't over-trigger on static content
- [x] `tests/apply-edit-dispatcher-component-extract.test.ts` — end-to-end: both files' on-disk content after a successful extract, `newFile` in the reply, stale refusal touches neither file, already-exists refusal, dry-run refusal, dynamic-expression refusal touches neither file, second-write-failure rollback (isolated via separate directories + chmod), and a real-git-repo test proving one commit covers both files and a single `undo_edit` reverts the whole extraction
- [x] `tests/apply-edit-schema-extract.test.ts` — schema validation for the new op/field, additive `newFile` on `createEditAppliedContent`
- [x] `test/edit-history.test.js` / `test/undo-edit.test.js` — new cases for `recordEdit`'s multi-file mode and `undoEdit`'s new-file-deletion path

**Update (follow-up commit `d0ae91a`):** an independent review caught that dynamic content (`{expression}` interpolation, dynamic attrs, spreads, template literals) inside the extracted subtree was silently copied into the new file instead of being refused — see "Dynamic content is refused, not copied" above. Fixed with `findDynamicContent` + 4 new tests; full suite re-verified green.

### DIFF
--- a/server/apply-edit-dispatcher.mjs
+++ b/server/apply-edit-dispatcher.mjs
@@ -5,8 +5,13 @@
  *   1. For `replace-image-src` with a raw `{filename, mimeType, dataURL}` value: pre-process
  *      via `processImageDrop` — write the bytes under `public/images/`, run optimize, build the
  *      `{src, srcset}` value the patcher's resolver expects. Throws map to `image-optimize-failed`.
- *   2. `resolve(projectRoot, edit)` (patcher.mjs) returns either `{file, range, replacement}`
- *      or `{refused: true, reason, detail?}`.
+ *   2. `resolve(projectRoot, edit)` (patcher.mjs) returns one of:
+ *        - `{file, range, replacement}` — the single-file case every op except extract-component
+ *          resolves to.
+ *        - `{extract: true, newFile: {path, content}, original: {file, range, replacement}}` —
+ *          extract-component's two-file case (component-extract-edit.mjs), handled by
+ *          `applyExtractComponent` below instead of the generic single-splice path.
+ *        - `{refused: true, reason, detail?}`.
  *   3. On refusal: forward as an `edit-failed` MCP response.
  *   4. On success: read the source, splice in the replacement, atomically write back (write to
  *      a sibling temp file then `rename`), invoke an optional `onApplied` hook so #298's
@@ -45,8 +50,8 @@ function failed(id, reason, detail) {
   return { content: [createEditFailedContent(id, reason, detail)], isError: true };
 }
 
-function applied(id, file, range, commit, result, model) {
-  return { content: [createEditAppliedContent(id, file, range, commit, result, model)] };
+function applied(id, file, range, commit, result, model, newFile) {
+  return { content: [createEditAppliedContent(id, file, range, commit, result, model, newFile)] };
 }
 
 /** Splice `replacement` into `source` at the resolved byte range. */
@@ -179,12 +184,99 @@ async function processImageDrop(projectRoot, edit) {
 }
 
 /**
+ * Handle extract-component's two-file write. `resolution` is component-extract-edit.mjs's
+ * `{extract: true, newFile: {path, content}, original: {file, range, replacement}}`.
+ *
+ * Design decision — write order and partial-failure handling: the brand-new component file is
+ * written FIRST, then the source file is patched second. This ordering makes the failure-of-the-
+ * second-write case cheap and safe to roll back: the new file never existed before this op, so
+ * "undo the first write" is just deleting it. The reverse order (patch source first, write new
+ * file second) would risk leaving the source file referencing a component
+ * (`<NewName ... />` + its import) that doesn't exist on disk if the second write then failed —
+ * a broken build, and a harder state to describe as "rolled back." There's still no cross-file
+ * transaction (this repo's `atomicWrite` only gives single-file atomicity via temp-sibling +
+ * rename — same accepted limitation `create-content.mjs`'s writes have), but new-file-first
+ * plus a best-effort unlink on second-write failure closes the realistic failure window: the
+ * only way to land in a truly inconsistent state is the unlink itself failing (e.g. permissions
+ * changed between the two writes), which is reported via `write-failed` either way.
+ *
+ * Once both writes succeed, both files are committed onto ONE `anglesite/edits` commit via
+ * `opts.onApplied({files: [...]})` (see edit-history.mjs's multi-file `recordEdit` mode) — so
+ * "one semantic op → one commit" holds here too, and a single `undo_edit` call reverts the whole
+ * extraction (both the new file's removal and the source file's restoration) atomically.
+ */
+async function applyExtractComponent(projectRoot, edit, resolution, opts) {
+  const { newFile, original } = resolution;
+  const absNewFile = join(projectRoot, newFile.path);
+  const absOriginal = join(projectRoot, original.file);
+
+  // Re-validate staleness against a FRESH read, immediately before writing — same TOCTOU
+  // discipline the generic single-file path applies below for every other COMPONENT_OPS member
+  // (component-extract-edit.mjs's own `await parse(...)` opens the same async yield point a
+  // concurrent edit could land in).
+  let freshSource;
+  try {
+    freshSource = readFileSync(absOriginal, "utf-8");
+  } catch (err) {
+    return failed(edit.id, "write-failed", `read ${original.file}: ${err.message}`);
+  }
+  if (fileVersion(freshSource) !== edit.component.baseVersion) {
+    return failed(edit.id, "stale", `${original.file} changed since the model was fetched`);
+  }
+  // The resolver already checked this once; re-check immediately before writing to narrow the
+  // race window against a concurrent extract targeting the same new name.
+  if (existsSync(absNewFile)) {
+    return failed(edit.id, "already-exists", `${newFile.path} already exists`);
+  }
+
+  try {
+    mkdirSync(dirname(absNewFile), { recursive: true });
+    atomicWrite(absNewFile, newFile.content);
+  } catch (err) {
+    return failed(edit.id, "write-failed", `${newFile.path}: ${err.message}`);
+  }
+
+  const next = spliceSource(freshSource, original.range, original.replacement);
+  try {
+    atomicWrite(absOriginal, next);
+  } catch (err) {
+    // Roll back the new file — see the design note above: new-file-first ordering makes this a
+    // plain unlink of something that never existed before this op, not a content restore.
+    try { unlinkSync(absNewFile); } catch {} // best-effort
+    return failed(edit.id, "write-failed", `${original.file}: ${err.message}`);
+  }
+
+  let commit;
+  if (opts.onApplied) {
+    try {
+      commit = await opts.onApplied({
+        files: [newFile.path, original.file],
+        projectRoot,
+        message: `anglesite: extract ${newFile.path} from ${original.file}`,
+      });
+    } catch {
+      commit = undefined; // patch landed; history-keeping failure shouldn't fail an already-applied edit
+    }
+  }
+
+  let model;
+  try {
+    model = await buildComponentModel(projectRoot, edit.component.path);
+  } catch {
+    model = undefined;
+  }
+
+  return applied(edit.id, original.file, original.range, commit, undefined, model, newFile.path);
+}
+
+/**
  * Apply an edit. Returns an MCP tool response (`{content, isError?}`) whose first content entry
  * is the JSON-encoded `edit-applied` / `edit-failed` body.
  *
  * @param {string} projectRoot
  * @param {object} edit  payload from the `apply_edit` tool (already zod-validated)
- * @param {{ onApplied?: (info: {file:string, range:{start:number,end:number}, projectRoot:string}) => Promise<string|undefined> | string | undefined }} [opts]
+ * @param {{ onApplied?: (info: {file:string, range:{start:number,end:number}, projectRoot:string}
+ *   | {files:string[], projectRoot:string, message?:string}) => Promise<string|undefined> | string | undefined }} [opts]
  */
 export async function applyEdit(projectRoot, edit, opts = {}) {
   // The app's Foundation Models chat path (ApplyEditTool, #251) forwards NL instructions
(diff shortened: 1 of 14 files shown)


server/component-extract-edit.mjs
```diff
@@ -0,0 +1,355 @@
...
+    if (n.kind === "text") {
+      const text = n.text; // trimmed by buildTemplateNodeIndex; may be silently truncated at 80 chars
+      if (text && text.length < 80 && n.span[0] != null && n.span[1] != null) {
+        const raw = source.slice(n.span[0], n.span[1]);
+        const trimmed = raw.trim();
+        if (trimmed === text) {
+          const start = n.span[0] + (raw.length - raw.trimStart().length);
+          const end = start + trimmed.length;
+          if (start >= nodeSpan[0] && end <= nodeSpan[1]) {
+            candidates.push({ value: text, kind: "text", span: [start, end] });
```
davidwkeith, line 174: **Confirmed bug — double-escapes text content that already contains HTML entities, verified with a live repro.**

`@astrojs/compiler` leaves text-node `.value` **raw/undecoded** (entities like `&amp;` are kept literally), but decodes attribute `.value` (single-pass). This candidate carries the raw text straight through into `instanceAttrs[].value`, which is later run through `escapeAttr()` (line 334) — and `escapeAttr`'s own doc comment says it expects a plain, decoded string ("escape `&` first so it doesn't double-escape the entities *this introduces*").

Repro (ran `resolveComponentExtract` directly against the PR's own files):

```astro
<!-- source -->
<h2>Fish &amp; Chips</h2>
```
extracting the `<h2>` produces this instance tag in the rewritten source:
```astro
<FishHeading text1="Fish &amp;amp; Chips" />
```
That's a corrupted, double-encoded value — on next parse it decodes once to the literal string `Fish &amp; Chips` (not `Fish & Chips`), which then gets re-escaped again by Astro's `{text1}` interpolation on render, so the page visibly shows `Fish &amp; Chips` instead of `Fish & Chips`. This directly contradicts the file's own doc comment that "the extraction is behavior-preserving by default" (line ~399).

Any real content authored with entities (very common — "Smith & Sons", "Terms & Conditions", imported/converted copy, etc.) hits this. No test in the PR covers entity-bearing text, so it ships uncaught.

Attribute-kind candidates (line 143, `attr.value`) are *not* affected — the compiler decodes those once, and `escapeAttr` re-encodes them once, so that round-trip is correct. The bug is specific to the raw/undecoded text-node path.

Fix: decode HTML entities in `c.value` before treating it as the "plain" value for `instanceAttrs`/prop default when `c.kind === "text"` (or otherwise normalize text vs. attr candidates to the same representation before they reach `escapeAttr`).

---

server/component-extract-edit.mjs
```diff
@@ -0,0 +1,355 @@
...
+  const fmLines = fmParts.join("\n");
+  const newFileContent = fmLines ? `---\n${fmLines}\n---\n${markup}\n` : `---\n---\n${markup}\n`;
+
+  const attrsText = instanceAttrs.map(({ name, value }) => ` ${name}="${escapeAttr(value)}"`).join("");
+  const instanceMarkup = `<${newName}${attrsText} />`;
+  const withInstance = source.slice(0, nodeSpan[0]) + instanceMarkup + source.slice(nodeSpan[1]);
+
+  const fmMatch = withInstance.match(/^(---\r?\n)([\s\S]*?)(\r?\n---)/);
+  let rewrittenOriginal;
+  if (!fmMatch) {
```
davidwkeith, line 387: Not a bug newly introduced by this PR (I verified the identical root cause already reproduces on `main` via `insert-node`'s equivalent fallback), but worth flagging since `extract-component` is a new path that hits the same trap: `/^(---\r?\n)([\s\S]*?)(\r?\n---)/` doesn't match a genuinely **empty** frontmatter (`---\n---\n`) — try it: `"---\n---\n<div/>".match(/^(---\r?\n)([\s\S]*?)(\r?\n---)/)` → `null`. So `fmMatch` is falsy even though frontmatter *does* exist, and this `!fmMatch` branch wrongly assumes there's none and prepends a brand-new `---\n...\n---\n` on top of the source's real (empty) fences.

Live repro against this PR's code, extracting from a component whose frontmatter is just `---\n---\n` (a completely ordinary "no props, no imports" presentational component — a very plausible first extraction target):

```
---
import FishHeading from "./FishHeading.astro";
---
---
---
<div>
  <FishHeading text1="..." />
</div>
```

Astro will parse only the first `---...---` block as frontmatter and render the second `---\n---\n` pair as literal text in the page — visible corruption, not a build break, but still wrong output.

Since the shared regex is what's actually broken, the cleanest fix is probably at that level (e.g. anchor on the *first* `---` line and the *next* `---`-only line rather than requiring a preceding blank line) so both `insert-node` and `extract-component` get fixed together — may be worth its own follow-up issue rather than blocking this PR, but calling it out since it's exercised right next to this new code.

---

server/component-extract-edit.mjs
```diff
@@ -0,0 +1,355 @@
...
+ *    generic `text1`, `text2`, … sequence in document order. Every hoisted prop is `type:
+ *    "string"` — no number/boolean inference (no prior art for it in this codebase, and the
+ *    payoff isn't worth the complexity for a first cut). Name collisions get a numeric suffix.
+ */
+
+function refuse(reason, detail) {
+  return { refused: true, reason, detail };
+}
+
+function validPath(relPath) {
```
davidwkeith, line 69: `validPath` is now duplicated a third time (byte-identical copies already exist in `component-structure-edit.mjs` and `component-frontmatter-edit.mjs`). This slightly undercuts the PR's own stated rationale for exporting `resolveAllSpans`/`escapeAttr`/`importSpecifier`/etc. from `component-structure-edit.mjs` — "reuse this module's single, carefully-tested implementation rather than a second, drift-prone copy." Worth exporting and importing this one too rather than tripling it.

---

server/apply-edit-dispatcher.mjs
```diff
@@ -178,13 +183,100 @@ async function processImageDrop(projectRoot, edit) {
...
+    return failed(edit.id, "stale", `${original.file} changed since the model was fetched`);
+  }
+  // The resolver already checked this once; re-check immediately before writing to narrow the
+  // race window against a concurrent extract targeting the same new name.
+  if (existsSync(absNewFile)) {
+    return failed(edit.id, "already-exists", `${newFile.path} already exists`);
+  }
+
+  try {
+    mkdirSync(dirname(absNewFile), { recursive: true });
```
davidwkeith, line 233: Minor: `mkdirSync(dirname(absNewFile), { recursive: true })` can create new directories before the write; if the source-file write below then fails, only `unlinkSync(absNewFile)` runs on rollback (line 245) — any directory this created is left behind. Low real-world impact since `newRelPath` is always `src/components/<newName>.astro`, which almost always pre-exists in a scaffolded site, but worth a note or a matching best-effort `rmdir` if you want the rollback comment's "closes the realistic failure window" claim to be fully accurate.

---

davidwkeith (commented):
Reviewed the `extract-component` two-file write/rollback/commit/undo machinery — it's careful and well-tested (stale checks, already-exists races, dry-run refusal, rollback-on-second-write-failure, and the `undo_edit` new-file-deletion path are all covered). No path-traversal risk (`newName` is constrained by `^[A-Z][A-Za-z0-9]*$` before touching the filesystem), and `newFile` on `createEditAppliedContent` is genuinely additive — confirmed no existing field changed.

One correctness bug should be fixed before merge: extracting a text node whose content already contains an HTML entity (e.g. `Fish &amp; Chips`, a very ordinary occurrence in real site copy) produces a double-escaped, corrupted attribute value on the new component instance — verified end-to-end by running `resolveComponentExtract` directly against the PR's own code (see inline comment on `component-extract-edit.mjs:133`). No test in the PR exercises entity-bearing content, so this ships silently.

I also found and verified (via direct repro) that extracting from a component with a genuinely *empty* frontmatter (`---\n---\n`) produces malformed doubled frontmatter fences — but that's a pre-existing bug in the shared `fmMatch` regex (I reproduced the identical failure on `main` via `insert-node`'s equivalent fallback), not something this PR introduces. Flagged inline for visibility since `extract-component` exercises the same trap, but not blocking on it here.

Two minor nits inline (a third duplicate of `validPath`, and an unrolled-back `mkdirSync` on the rollback path) — not blocking.

---

PR #421, opened 2026-07-17
feat(mcp): add extract-component apply_edit op (Component Editor Slice 5)

davidwkeith:
## Summary

- Adds `extract-component` to the `apply_edit` MCP tool — the first op that writes **two files** atomically in one edit: carves a `get_component_model` outline subtree (element/component/slot only) out of an existing `.astro` component into a brand-new file under `src/components/`, and replaces the extracted markup with a self-closing instance + import.
- Hoists the original component's own declared Props that the subtree bare-referenced (`props-interface.mjs`'s `parseProps`), copies/rebases/prunes nested-component imports carried into the moved markup, and migrates simple-selector (tag/class/id only) scoped-style rules that exclusively target the extracted markup — anything shared or too complex to analyze automatically stays behind with a non-blocking entry in a new `warnings` array.
- Both files land in **one** commit on the hidden `anglesite/edits` history branch, so `undo_edit` reverts both together in a single step.
- `server/undo-edit.mjs` gains a fix for its own latent gap: it previously assumed every touched file existed in the parent commit (true for every other op) and crashed reverting a brand-new file. It now deletes files that didn't exist at the parent commit instead of trying to restore them.

Design: `docs/superpowers/specs/2026-07-16-extract-component-design.md`
Plan: `docs/superpowers/plans/2026-07-16-extract-component.md`

## Paired PR check

- [ ] This change **needs a paired PR** in [`Anglesite/Anglesite-app`](https://github.com/Anglesite/Anglesite-app) — the app-side "Extract into Component…" UI, name-prompt dialog, and `result.warnings`/`hoistedProps` surfacing (Anglesite-app#495, Component Editor Slice 5). Not included in this PR; app-side work is a separate follow-up per the established Slice 4 pattern (PR #418 / Anglesite-app#494).
- [ ] Version bump (`bin/release.ts`) is intentionally **not** included here — versioning/tagging/release should happen from `main` after this branch merges, not from the feature branch.

## Test plan

- [x] `npx vitest run` — 3086 passed, 1 pre-existing unrelated failure (`test/optimize-images-packaging.test.js`, missing local `node_modules/.bin/tsx` binary — confirmed via `npm ls tsx` returning empty; last touched by an unrelated PR #402, not caused by this branch).
- [x] New tests across 6 files: `tests/style-selector-match.test.ts`, `tests/component-extract-edit.test.ts` (22 cases covering the core move, prop hoisting, nested-import copy/prune, and style migration), `tests/apply-edit-schema-extract.test.ts`, `tests/apply-edit-dispatcher-component-extract.test.ts` (full round trip, `exists` refusal, `dry_run`, one-commit-undo-reverts-both, concurrency/staleness race), plus additions to `test/edit-history.test.js` and `test/undo-edit.test.js`.
- [x] Built via 10 TDD subagent-driven tasks, each independently spec-compliance + code-quality reviewed (several real bugs — empty-frontmatter corruption, duplicate nested-component imports, a stale-span rule-matching bug, the `undo_edit` new-file gap — were caught and fixed during those reviews), plus a final whole-branch review (**Ready to merge: Yes**, no Critical/Important issues).

### DIFF
--- a/server/apply-edit-dispatcher.mjs
+++ b/server/apply-edit-dispatcher.mjs
@@ -6,13 +6,16 @@
  *      via `processImageDrop` — write the bytes under `public/images/`, run optimize, build the
  *      `{src, srcset}` value the patcher's resolver expects. Throws map to `image-optimize-failed`.
  *   2. `resolve(projectRoot, edit)` (patcher.mjs) returns either `{file, range, replacement}`
- *      or `{refused: true, reason, detail?}`.
+ *      (plus, for `extract-component`, a second `newFile: {path, content}`) or
+ *      `{refused: true, reason, detail?}`.
  *   3. On refusal: forward as an `edit-failed` MCP response.
- *   4. On success: read the source, splice in the replacement, atomically write back (write to
- *      a sibling temp file then `rename`), invoke an optional `onApplied` hook so #298's
- *      `edit-history.mjs` can commit to the hidden `anglesite/edits` branch and thread its SHA
- *      back as `commit`, then return `edit-applied` — with `result: {src, srcset}` for the
- *      image-drop path.
+ *   4. On success: for `extract-component`, create-write `newFile` first (fails as `exists` on
+ *      EEXIST, so the original is never touched if the new path is already taken) — then read
+ *      the source, splice in the replacement, atomically write back (write to a sibling temp
+ *      file then `rename`), invoke an optional `onApplied` hook so #298's `edit-history.mjs`
+ *      can commit both files to the hidden `anglesite/edits` branch in one commit and thread its
+ *      SHA back as `commit`, then return `edit-applied` — with `result: {src, srcset}` for the
+ *      image-drop path, or `result: {componentPath, hoistedProps, warnings}` for extract-component.
  *   5. On any filesystem error: return `edit-failed` with reason `write-failed`.
  *
  * The handler stays a pure async function so it's unit-testable independent of the MCP
@@ -69,8 +72,8 @@ function windowAround(source, next, pad = 200) {
   return { before: source.slice(from, beforeTo), after: next.slice(from, afterTo) };
 }
 
-function preview(id, file, range, op, before, after) {
-  return { content: [createEditPreviewContent(id, file, range, op, before, after)] };
+function preview(id, file, range, op, before, after, newFile) {
+  return { content: [createEditPreviewContent(id, file, range, op, before, after, newFile)] };
 }
 
 /** Atomic write via temp-sibling + rename, so a crashed mid-write can't truncate the source. */
@@ -184,7 +187,7 @@ async function processImageDrop(projectRoot, edit) {
  *
  * @param {string} projectRoot
  * @param {object} edit  payload from the `apply_edit` tool (already zod-validated)
- * @param {{ onApplied?: (info: {file:string, range:{start:number,end:number}, projectRoot:string}) => Promise<string|undefined> | string | undefined }} [opts]
+ * @param {{ onApplied?: (info: {file:string, range:{start:number,end:number}, projectRoot:string, newFile?:{path:string}}) => Promise<string|undefined> | string | undefined }} [opts]
  */
 export async function applyEdit(projectRoot, edit, opts = {}) {
   // The app's Foundation Models chat path (ApplyEditTool, #251) forwards NL instructions
@@ -251,7 +254,27 @@ export async function applyEdit(projectRoot, edit, opts = {}) {
 
   if (edit.dry_run) {
     const { before, after } = windowAround(source, next);
-    return preview(edit.id, file, range, edit.op, before, after);
+    const newFilePreview = resolution.newFile
+      ? { path: resolution.newFile.path, after: resolution.newFile.content }
+      : undefined;
+    return preview(edit.id, file, range, edit.op, before, after, newFilePreview);
+  }
+
+  // extract-component writes a second, brand-new file. Do that FIRST with a create-if-absent
+  // flag ("wx" — fails on EEXIST) so the only possible partial-failure state is "new file
+  // exists, original untouched" — never the reverse. If this fails, the primary atomicWrite
+  // below never runs, so the original component is guaranteed unmodified.
+  if (resolution.newFile) {
+    const absNewPath = join(projectRoot, resolution.newFile.path);
+    try {
+      mkdirSync(dirname(absNewPath), { recursive: true });
+      writeFileSync(absNewPath, resolution.newFile.content, { encoding: "utf-8", flag: "wx" });
+    } catch (err) {
+      if (err.code === "EEXIST") {
+        return failed(edit.id, "exists", `${resolution.newFile.path} already exists`);
+      }
+      return failed(edit.id, "write-failed", `${resolution.newFile.path}: ${err.message}`);
+    }
   }
 
   try {
@@ -263,7 +286,12 @@ export async function applyEdit(projectRoot, edit, opts = {}) {
   let commit;
   if (opts.onApplied) {
     try {
-      commit = await opts.onApplied({ file, range, projectRoot });
+      commit = await opts.onApplied({
+        file,
+        range,
+        projectRoot,
+        newFile: resolution.newFile ? { path: resolution.newFile.path } : undefined,
+      });
     } catch (err) {
       // Patch landed on disk but history-keeping failed. Surface as a successful apply with no
       // commit SHA — the user-visible source change is real; #298 can decide its own policy.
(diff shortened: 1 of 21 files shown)


server/component-extract-edit.mjs
```diff
@@ -0,0 +1,359 @@
...
+// `findFrontmatterBody` is the sole frontmatter detector in this module (see its file-level
+// comment above) — it correctly matches the minimal empty-body frontmatter shape
+// "---\n---\n", which a regex requiring two independently-matched newlines would miss,
+// causing a second frontmatter block to be prepended via the no-frontmatter fallback below
+// and corrupting the file.
+function rewriteOriginalFrontmatter(afterNode, relPath, componentName, newComponentPath, movedComponentNames) {
+  const specifier = importSpecifier(relPath, newComponentPath);
+  const fm = findFrontmatterBody(afterNode);
+  const ensureAndPrune = (fmBody) => {
+    let body = ensureImport(fmBody, { localName: componentName, specifier }).source;
```
davidwkeith, line 152: **Bug: duplicate/conflicting import when the new component's name collides with an existing import's local name under a different specifier.**

`ensureImport` (`server/frontmatter-imports.mjs:19-36`) dedups only by `specifier`, never by `localName`:

```js
if (imports.some((i) => i.specifier === specifier)) {
  return { source: frontmatterSource, added: false };
}
```

So if the source file already has, say, `import Hero from "../shared/OtherHero.astro";` and the user extracts a subtree into `src/components/Hero.astro` (i.e. `componentName === "Hero"`), nothing here refuses the collision — `ensureAndPrune` appends a second `import Hero from "./Hero.astro";`. I reproduced this against this branch: the result is two conflicting `import Hero` declarations in the same frontmatter (invalid ESM — duplicate identifier), plus every existing `<Hero />` usage in the file becomes ambiguous between the two imports.

The design doc calls for refusing on this kind of collision rather than guessing an alternate name, and the resolver already has an `exists` refusal path for `newComponentPath` collisions (around line 299) — but there's no equivalent check here for `componentName` colliding with an existing import's `localName` (under a different specifier). Worth adding a check in `rewriteOriginalFrontmatter` (or earlier, alongside the `newComponentPath` exists-check) that inspects `parseImports(fmBody)` for a same-name/different-specifier entry and refuses with `invalid-input`, matching the existing collision-refusal pattern.

No test in `tests/component-extract-edit.test.ts` currently exercises this case — every existing "collision" test is for `newComponentPath` already existing on disk, not for an import-name collision.

(Minor, related: when the frontmatter's last import sits directly against the closing `---` fence with no trailing newline in `fmBody`, `ensureImport`'s insertion glues the new import onto the previous line with no line break — e.g. `import Badge from "./Badge.astro";import Hero from "./Hero.astro";`. Syntactically valid but worth a newline guard; it's what makes the repro above extra ugly to read.)

---

server/undo-edit.mjs
```diff
@@ -76,10 +76,22 @@ export async function undoEdit(projectRoot, { commit, force = false } = {}) {
...
 
-  // 6. Write parent's blob content for each file back to disk.
+  // 6. Restore each touched file: files that existed at `parent` get their old blob content
+  // written back; files that were pure additions at `head` (didn't exist at `parent` — e.g. the
+  // new component file extract-component creates) get deleted instead, since "undo" for a newly
+  // created file means removing it.
   for (const file of files) {
-    const content = runGitBuffer(projectRoot, ["show", `${parent}:${file}`]);
-    writeFileSync(join(projectRoot, file), content);
+    const existedAtParent = tryRunGit(projectRoot, ["cat-file", "-e", `${parent}:${file}`]) !== undefined;
```
davidwkeith, line 84: **Risk: this can't distinguish "file didn't exist at parent" from "git failed for some unrelated reason," and the unrelated-failure case silently deletes data instead of restoring it.**

```js
const existedAtParent = tryRunGit(projectRoot, ["cat-file", "-e", `${parent}:${file}`]) !== undefined;
```

`tryRunGit` swallows every error uniformly:

```js
function tryRunGit(projectRoot, args, env = {}) {
  try { return runGit(projectRoot, args, env); } catch { return undefined; }
}
```

So a transient object-store read failure, permissions issue, or any other `cat-file -e` failure looks identical to "genuinely absent at `parent`." In that misclassified case, the code takes the `else` branch below and `unlinkSync`s the file instead of restoring the parent's blob — then still reports `status: "undone"` (success), with nothing indicating content was discarded rather than restored.

Pre-PR, the equivalent restore step threw on any such git error and surfaced as a failure. This new branch trades that fail-safe behavior for silent data loss on the (narrow, but real) unrelated-error path. Given `unlinkSync` is destructive, it'd be safer to have `tryRunGit` here (or a dedicated variant) distinguish "process exited because the path is absent" (e.g. check `err.status === 1` and stderr shape for `cat-file -e`) from any other failure, and surface the latter as a refusal rather than proceeding.

Also untested — `test/undo-edit.test.js`'s new case only covers the clean "genuinely new file" and "genuinely pre-existing file" happy paths, not a `cat-file -e` failure for an unrelated reason.

---

server/apply-edit-dispatcher.mjs
```diff
@@ -251,7 +254,27 @@ export async function applyEdit(projectRoot, edit, opts = {}) {
...
+      ? { path: resolution.newFile.path, after: resolution.newFile.content }
+      : undefined;
+    return preview(edit.id, file, range, edit.op, before, after, newFilePreview);
+  }
+
+  // extract-component writes a second, brand-new file. Do that FIRST with a create-if-absent
+  // flag ("wx" — fails on EEXIST) so the only possible partial-failure state is "new file
+  // exists, original untouched" — never the reverse. If this fails, the primary atomicWrite
+  // below never runs, so the original component is guaranteed unmodified.
+  if (resolution.newFile) {
```
davidwkeith, line 267: Accepted-by-design partial-failure state, but currently unhandled: a few lines below this block, `atomicWrite(absPath, next)` patches the original file. If that throws *after* `resolution.newFile` was already written here, there's no cleanup of the just-created file — `unlinkSync` in this file only appears inside `atomicWrite`'s own temp-file rollback, never applied to `resolution.newFile`'s path on this path. The design doc accepts "new file exists, original untouched" as the only reachable partial-failure state, which is true, but two things follow from not cleaning it up:

1. An orphaned, untracked file is left in the user's working tree indefinitely after a transient failure (e.g. disk full, permissions).
2. A caller retrying the identical `extract-component` call gets a confusing `exists` refusal (from the `wx` flag above) instead of the retry succeeding — the op becomes unrecoverable without manually deleting the orphaned file first.

Worth a best-effort `unlinkSync(absNewPath)` in the `atomicWrite` catch below, mirroring the rollback pattern `atomicWrite` already uses on its own temp file, so a failed apply leaves no residue and is safely retryable.

---

davidwkeith (commented):
Reviewed the `extract-component` implementation (span/offset arithmetic, two-file write ordering + rollback, `edit-history.mjs`'s multi-file commit staging, `undo_edit`'s new-file handling, prop hoisting, import hygiene, and style migration). Most of it holds up well and is backed by real tests — the subtree-root span resolution, the write-ordering guarantee (new file never orphans a broken reference in the source), the multi-file git staging, and the style-selector complexity check are all sound.

Three issues worth addressing before merge, most severe first (GitHub won't let me use "Request changes" on my own PR, but treat finding #1 as blocking):

1. **`server/component-extract-edit.mjs`** — confirmed, reproducible bug: extracting into a component name that collides with an existing import's local name (different specifier) produces two conflicting `import X` declarations in the source file's frontmatter — invalid ESM, and every existing usage of that name becomes ambiguous. `ensureImport` only dedups by specifier, not local name, and there's no separate collision check for this case (unlike the existing `newComponentPath`-already-exists check). See inline comment for a full repro.
2. **`server/undo-edit.mjs`** — the new "was this file a pure addition?" check via `cat-file -e` can't distinguish genuine non-existence from an unrelated git failure (`tryRunGit` swallows all errors uniformly), and the unrelated-failure case silently deletes the file instead of restoring it, while still reporting success.
3. **`server/apply-edit-dispatcher.mjs`** — no rollback of the newly-written component file if the subsequent primary-file write fails; leaves an orphaned untracked file and makes retry fail with a confusing `exists` error instead of succeeding.

None of these are hit by the current test suite, which is otherwise pretty thorough for this feature. Happy to re-review once addressed.

---

davidwkeith:
Closing as superseded by #420, which landed a different (and already-released, v1.7.0) implementation of the same `extract-component` op while this PR was in review.

The two designs are materially incompatible (payload field `newComponentPath` vs `newName`, response shape `result: {componentPath, hoistedProps, warnings}` vs a top-level `newFile: string`, different prop-hoisting heuristic, and #420 refuses on dynamic content rather than partially hoisting it), so reconciling them isn't a simple merge — #420 is the accepted implementation going forward.

One thing this branch has that #420 doesn't: simple-selector scoped-style-rule migration for the extracted markup. Not carrying that over here — flagging it as a possible small follow-up on top of main if it's still wanted, rather than resurrecting this branch.

The paired app-side PR (Anglesite/Anglesite-app#805) is being reworked against #420's actual contract.

PR #423, opened 2026-07-25
docs: trim derivable content from root CLAUDE.md

davidwkeith:
## Summary

- Cut ~two-thirds of the root `CLAUDE.md` — content a session can reconstruct by reading the repo. The file loads in full every session, so it went from **28,632 → ~10,300 chars (~4,600 est. tokens saved per session)**.
- Removed as derivable: the 141-line annotated directory tree (`ls` shows it), both 56-row skills reference tables (they duplicated each skill's own frontmatter `description` *and* the generated `docs/dev/skill-registry.md`), the `Testing` section (vitest version + npm scripts + dir layout all live in `package.json`/`vitest.config.ts`), and the CI/CD job-by-job walkthrough (both workflow files are readable). Also dropped a stale ADR count: it claimed 24, `docs/decisions/` holds 25.
- Moved the optional Serena setup into `.claude/skills/serena/SKILL.md` so only its one-line description stays resident.
- Regenerated the `agent-skills/` export, which was stale on `main` and blocking all PRs (see below).

Deliberately preserved, because the codebase can't teach them: the `content-types.mjs` ↔ `ContentTypeRegistry.swift` source-of-truth rule (now a **Typed content** note in the MCP section), the user-facing/model-only frontmatter convention, and the directive to track the app's pinned Node version in the test matrix. Also kept whole: the agent instruction hierarchy table, the MCP/Anglesite-app contract, every editing guideline (cross-platform bans, PII-on-demand, MCP-first provisioning), key decisions, version management, distribution channels, the non-guessable manual scaffold recipe, and the security-hooks policy.

`template/CLAUDE.md` is untouched — it's a product artifact copied into users' projects by `scripts/scaffold.sh`, not developer memory.

### Correction: the `**Version:**` line stays

The first pass cut it as a redundant fifth copy of an already-thrice-synced number. That was wrong, and CI caught it (`5cee9a6` restores it). The line is machine-maintained: `bin/release.ts:59` **throws** if it's missing, so removing it would have broken the next release outright, and `tests/release.test.ts:100` asserts it matches `package.json`. It's back, now carrying a note recording that constraint so the same heuristic isn't reapplied. The license / Node / module-system fields on that line remain dropped — those are genuinely derivable.

### Pre-existing on `main`: stale `agent-skills/` export

`agent-skills/` was stamped `1.6.0` while every manifest reads `1.7.0`, so `agent-skills-export` and 56 `build-agent-skills` assertions failed on `main` independently of this PR. Since that check gates every PR, `5675b0a` regenerates it — pure `npm run build:agent-skills` output (version stamp in 56 `SKILL.md` files, two bundled reference `package.json` copies, one README line; no content changes), in its own commit to stay reviewable separately.

This is the second occurrence: `8bbfbe2` fixed it for 1.6.0, then `082c59c` ("chore: release v1.7.0") reintroduced it. Root cause is that `bin/release.ts` bumps manifests without regenerating derived files. Two related items left **out** of this PR: the root `package-lock.json` is stamped `1.5.0` (no CI job gates it — `template-lockfile` only checks `template/`), and wiring the export regen into `bin/release.ts` would end the recurrence. Both belong in a follow-up that changes the release path.

## Paired PR check

Docs plus a regenerated export. No MCP message schema, template field, or hook surface changes — the MCP section's tool table and transport contract are carried over verbatim, with one clarifying note added.

## Test plan

- [x] Plugin self-checks pass — full suite run locally: **3065 passed, 3 failed**
- [x] If touching the template: `scripts/scaffold.sh` produces a buildable site — n/a, `template/` untouched
- [x] If touching the MCP server: paired app PR exercises the new/changed messages — n/a, `server/` untouched

The 3 local failures are a sandbox artifact, not a regression: `apply-edit-dispatcher` "write should fail when the directory isn't writable" cases, which cannot fail under **uid 0** because root ignores permission bits. They are absent from CI's failure list, which runs non-root.

Suites covering the CI failures and every file touched here pass locally: `build-agent-skills` (87), `release` (11), `skill-registry` (24), `markdownlint` (5), `build-instructions` (21).

Verified by inspection as well:

- `bin/build-instructions.ts` validates `template/CLAUDE.md` only (lines 276, 358) — not the root file this PR edits.
- `tests/markdownlint.test.ts` lints `template/**/*.md`; its only root-`CLAUDE.md` mention is a code comment pointing at the "Testing changes manually" section, which this PR keeps.
- `bin/build-agent-skills.ts` globs `join(root, "skills")`, so `.claude/skills/serena/` can't leak into the export or trip the staleness check. That reasoning is now a comment at the glob itself rather than only here in the description, per review feedback.

### DIFF
--- a/.claude/skills/serena/SKILL.md
+++ b/.claude/skills/serena/SKILL.md
@@ -1,0 +1,20 @@
+---
+name: serena
+description: Set up Serena for semantic, symbol-level code navigation when working on the Anglesite plugin itself — indexing the project and starting its MCP server via uvx. Use when tracing cross-skill references or finding symbol usages across the skills tree.
+---
+
+# Serena (optional, plugin development)
+
+[Serena](https://github.com/oraios/serena) provides semantic, symbol-level code navigation via language servers. It's useful when working on the plugin itself (tracing cross-skill references, finding symbol usages across the skills tree). Not required — all standard tools work without it.
+
+**Setup:**
+
+```sh
+# Index the project (one-time)
+uvx -p 3.13 --from git+https://github.com/oraios/serena serena project index
+
+# Start the MCP server
+uvx -p 3.13 --from git+https://github.com/oraios/serena serena start-mcp-server --project .
+```
+
+Config lives in `.serena/project.yml`. Requires Python 3.13 and [uv](https://docs.astral.sh/uv/).
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -2,6 +2,10 @@
 
 Anglesite is a Claude plugin that scaffolds and manages websites for small businesses. It works with Claude Cowork (non-technical site owners, GUI) and Claude Code (developers, CLI). It generates Astro + Keystatic sites deployed to Cloudflare Workers (Static Assets, via the `@astrojs/cloudflare` adapter).
 
+**Version:** 1.7.0
+
+> Don't hand-edit or remove that line: `bin/release.ts` rewrites it on every bump and **throws** if it's missing, and `tests/release.test.ts` asserts it matches `package.json`.
+
 ## Agent instruction hierarchy
 
 Two levels of agent instructions exist — do not confuse them:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -2,150 +2,6 @@
 
 Anglesite is a Claude plugin that scaffolds and manages websites for small businesses. It works with Claude Cowork (non-technical site owners, GUI) and Claude Code (developers, CLI). It generates Astro + Keystatic sites deployed to Cloudflare Workers (Static Assets, via the `@astrojs/cloudflare` adapter).
 
-**Version:** 1.7.0 · **License:** ISC · **Node:** >=22 · **Module system:** ESM
-
-## Plugin structure
-
-```
-├── .claude-plugin/
-│   ├── plugin.json               Plugin manifest (name, version, metadata)
-│   └── marketplace.json          Marketplace catalog — this repo is also its own marketplace
-├── skills/                       Skills (56 total: 31 user-facing, 25 model-only)
-│   ├── start/SKILL.md            First-time setup + scaffolding
-│   ├── deploy/SKILL.md           Build, scan, deploy to Cloudflare Workers
-│   ├── check/SKILL.md            Health audit + troubleshooting
-│   ├── update/SKILL.md           Update dependencies + template files
-│   ├── domain/SKILL.md           DNS management (email, Bluesky, verification)
-│   ├── import/SKILL.md           Import from website URL
-│   ├── convert/SKILL.md          Convert existing SSG project to Anglesite
-│   ├── export/SKILL.md           Portable export (dist + content + MIGRATING.md)
-│   ├── contact/SKILL.md          Contact form (Workers + Turnstile)
-│   ├── forms/SKILL.md            Custom forms — RSVP, lead, survey, callback (Workers + Turnstile)
-│   ├── inbox/SKILL.md            Form submissions inbox in Keystatic (Cloudflare D1 + triage + CSV export)
-│   ├── indieweb/SKILL.md         Self-owned IndieAuth + Webmention + Micropub endpoints (user-facing)
-│   ├── backup/SKILL.md           Back up changes to GitHub
-│   ├── stats/SKILL.md            Plain-language site analytics
-│   ├── newsletter/SKILL.md       Email newsletter setup + subscribe form
-│   ├── add-store/SKILL.md        Ecommerce intake (user-facing)
-│   ├── design-interview/SKILL.md Visual identity (model-only)
-│   ├── email/SKILL.md            Business email setup, Apple-first (model-only)
-│   ├── animate/SKILL.md          CSS animations (model-only)
-│   ├── new-page/SKILL.md         Page creation (model-only)
-│   ├── repurpose/SKILL.md        Social media post generation (model-only)
-│   ├── seasonal/SKILL.md         Seasonal content suggestions (model-only)
-│   ├── optimize-images/SKILL.md  Image optimization pipeline (model-only)
-│   ├── og-images/SKILL.md       Satori-based OG image generation (model-only)
-│   ├── business-info/SKILL.md    Hours, location, LocalBusiness JSON-LD (model-only)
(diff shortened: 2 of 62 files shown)


davidwkeith (commented):
Reviewed the diff and fact-checked the PR description's claims against the repo:

- `docs/dev/skill-registry.md` is real, generated (`npm run registry`), 56/31/25 counts match, and `.github/workflows/test.yml` already fails CI if it goes stale — so pointing `CLAUDE.md`'s Skills reference section at it instead of a hand-maintained duplicate table is safe long-term, not just today.
- `bin/build-instructions.ts` only validates `template/CLAUDE.md` (confirmed lines 276/358) — untouched by this PR, as claimed.
- `bin/build-agent-skills.ts` globs `join(root, "skills")`, so the new `.claude/skills/serena/SKILL.md` correctly can't leak into the `agent-skills/` export or trip the staleness check.
- `docs/decisions/` does hold 25 entries (24 ADRs + README) — the deleted "ADR-0001 through ADR-0023" line was in fact stale (ADR-0024 exists), and the new text avoids re-introducing a count that'll drift again.
- The "Typed content" note wasn't invented — it's the one non-derivable fact from the deleted directory-tree comment (`content-types.mjs` ↔ `ContentTypeRegistry.swift`), correctly preserved as prose before the tree was cut.
- No dangling links/anchors into the removed sections (`## Plugin structure`, `## Testing`, `## Serena`) found elsewhere in the repo.
- `.claude/skills/serena/SKILL.md` is a sensible home for this — it's a Claude-Code-native project skill for people developing the plugin itself, distinct from `skills/` (which ships to end users via the marketplace/Open Agent Skills export). Frontmatter is well-formed.

Nothing to fix — this is a clean, well-scoped, docs-only trim, and the PR body's engineering rationale checks out against the actual source rather than just being asserted. (Note: I couldn't formally approve — GitHub reports this PR as authored by the same account I'm reviewing as.)

One optional nit for a future pass, not a blocker: the PR body's "Deliberately preserved" paragraph is now the only place explaining *why* `.claude/skills/serena/` won't leak into the agent-skills export — that reasoning lives in the PR description, not in a comment anywhere in the repo, so it'll be lost once this PR merges and the description scrolls out of easy reach. Just flagging it as institutional-memory risk if `bin/build-agent-skills.ts`'s glob ever needs to change.

---

davidwkeith:
CI came back red with two unrelated causes. One was my mistake; the other predates this PR. Both are fixed in `5cee9a6` and `5675b0a`.

**1. My error — I cut a line that isn't derivable (`5cee9a6`)**

`tests/release.test.ts:100` failed: `CLAUDE.md should contain a **Version:** line`. The trim reasoning was that the version line was a redundant fifth copy of a number already synced across three manifests. The duplication part was right; the conclusion was wrong. That line is machine-maintained:

- `bin/release.ts:59` **throws** `No **Version:** line found` if it's absent — so removing it wouldn't just fail a test, it would have broken the next `bin/release.ts` run outright.
- `tests/release.test.ts:100` asserts it equals `package.json`'s version.

Restored. I kept the license / Node / module-system fields dropped — those really are derivable from the manifests — and added a note recording the constraint, so the line doesn't read as stale prose to the next person applying the same heuristic:

```markdown
**Version:** 1.7.0

> Don't hand-edit or remove that line: `bin/release.ts` rewrites it on every bump
> and **throws** if it's missing, and `tests/release.test.ts` asserts it matches
> `package.json`.
```

Worth flagging that my "no dangling references" check was tool-assisted and missed this: I grepped for links and anchors into the removed sections, but a *field consumed by a build script* isn't a link. Your review independently confirmed the parts I did verify, and neither of us caught this one.

**2. Pre-existing on `main` — stale `agent-skills/` export (`5675b0a`)**

`agent-skills-export` plus 56 `build-agent-skills.test.ts:312` assertions failed because the committed export is stamped `1.6.0` while every manifest reads `1.7.0`. This is independent of this PR — `main` is red on it right now, and this branch never touched `agent-skills/`:

```
origin/main  package.json                              1.7.0
origin/main  agent-skills/update/references/package.json  1.6.0
origin/main  agent-skills/deploy/SKILL.md   version: "1.6.0"
```

Normally I'd leave a base-branch failure alone, but this check gates *every* PR, so nothing can merge until it's regenerated. It's pure `npm run build:agent-skills` output — the version stamp in 56 `SKILL.md` files, two bundled reference `package.json` copies, one README line, and no content changes:

```
56 -  version: "1.6.0"   →  56 +  version: "1.7.0"
```

It's in its own commit so it stays reviewable separately from the docs trim.

**This is the second occurrence.** `8bbfbe2` ("regenerate stale agent-skills export for 1.6.0 version bump") fixed exactly this, then `082c59c` ("chore: release v1.7.0") reintroduced it. The root cause is that `bin/release.ts` bumps the manifests and `CLAUDE.md` but never regenerates derived files — so it will happen again on 1.8.0.

Two things point at the same gap, which I've deliberately left out of this PR:

- **Root `package-lock.json` is stamped `1.5.0`** — two releases behind. My `npm install` corrected it as a side effect and I reverted that; no CI job gates the root lockfile (`template-lockfile` only checks `template/`), so it isn't blocking anything.
- **Wiring `npm run build:agent-skills` (and a lockfile refresh) into `bin/release.ts`** would end the recurrence.

Happy to open that as a follow-up if you want it — it's a behavior change to the release path, so it seemed wrong to bundle into a docs PR.

**3. Your review nit — addressed (`5cee9a6`)**

Fair catch that the "why `.claude/skills/` can't leak into the export" reasoning lived only in the PR description. I put it at the glob, where someone changing that line would actually see it, rather than in prose that has to be found:

```ts
// Only skills/ is exported — deliberately NOT .claude/skills/, which holds
// Claude-Code-native skills for developing this plugin (never shipped to end
// users). Widening this glob would publish internal tooling to skills.sh.
const skillsDir = join(root, "skills");
```

**Local verification** (`npm install` run this time, so the suite actually executes): the five suites covering the failures and the files I touched all pass — `build-agent-skills` (87), `release` (11), `skill-registry` (24), `markdownlint` (5), `build-instructions` (21).

Full suite is 3065 passed / 3 failed, and those 3 are a sandbox artifact, not a regression: `apply-edit-dispatcher` "write should fail when the directory isn't writable" cases, which can't fail for **uid 0** because root ignores permission bits. They're absent from CI's failure list, which runs non-root. Not touching them.

---

PR #424, opened 2026-07-27
docs(adr): record snapshotted embeds, reject Zaraz SSR

davidwkeith:
## Summary

- Adds **ADR-0025 — Snapshot social embeds into the owner's repo instead of rendering them remotely**, recording the evaluation asked for in [Anglesite-app#682](https://github.com/Anglesite/Anglesite-app/issues/682).
- Amends **ADR-0008**'s alternatives table: the social-embed row moves from "static screenshots with links to original posts" to snapshotted first-party embed cards, and the YouTube row from "link to video with a thumbnail image" to a snapshotted card with a self-hosted thumbnail.
- Regenerates `agent-skills/` so the 34 vendored copies of ADR-0008 match the canonical file — CI fails on a stale export.

## Paired PR check

- [ ] This change **needs a paired PR** in [`Anglesite/Anglesite-app`](https://github.com/Anglesite/Anglesite-app). Link it here:

Docs-only, no MCP schema change. The implementation it records is [Anglesite/Anglesite-app#979](https://github.com/Anglesite/Anglesite-app/pull/979), which does not depend on this PR landing — merge order doesn't matter.

## Test plan

- [x] Plugin self-checks pass — `npm run build:agent-skills` is clean and **idempotent** (a second run produces no further changes), so the committed export is not stale. Diff is confined to the 34 vendored `0008-no-third-party-javascript.md` copies plus the three intentional `docs/decisions/` changes; nothing unrelated moved.
- [ ] If touching the template: `scripts/scaffold.sh` produces a buildable site — n/a, no template change.
- [ ] If touching the MCP server: paired app PR exercises the new/changed messages — n/a, no MCP change.

## Why ADR-0008 needed amending, not just an exception

The point worth making explicit — and now stated in ADR-0008 itself — is that these two rows add **no exception to the list**. A snapshotted card is first-party HTML and first-party images served from the owner's own domain, so it needs no CSP allowlist entry at all. That is the test of whether an "alternative" genuinely preserves this ADR or quietly widens it, and it's why hotlinking the platform's images was rejected alongside the scripts: a hotlinked image still leaks every visitor's IP and `Referer` to the platform.

## Why Zaraz was rejected

Zaraz's server-side rendering of embeds genuinely does solve the third-party-JavaScript problem — Cloudflare fetches the post at the edge, strips the scripts, and proxies images through the owner's domain. On the narrow ADR-0008 question it is a valid answer, which is why it's recorded as a considered option rather than dismissed.

It was rejected because it solves the problem *at request time, at the edge*: the rendered artifact is invisible in `astro dev` and `preview`, evaporates if the site leaves Cloudflare, and is configured in a per-zone dashboard rather than the owner's repo — exactly the invisible, unportable configuration ADR-0011 and the git-is-the-source-of-truth rule exist to keep out. It also covers X and Instagram only, and cannot outlive the platform. Since the premise of the whole feature is that these platforms are decaying, that last point is decisive rather than incidental.

`astro-embed` was also considered and rejected (fetches on every build, hotlinks media, and a deleted upstream post silently empties the embed), and the app PR removes it as an unused dependency.

## Note on cross-references

ADR-0008's new "(see ADR-0025)" mentions are not bundled into the skills that vendor ADR-0008 — the generator bundles only the ADRs each `SKILL.md` explicitly links. That matches the existing shipped state, where ADR-0008's own "(see ADR-0005)" is likewise an unbundled cross-reference.

### DIFF
--- a/docs/decisions/0025-snapshotted-embeds.md
+++ b/docs/decisions/0025-snapshotted-embeds.md
@@ -1,0 +1,147 @@
+---
+status: accepted
+date: 2026-07-26
+decision-makers: [Anglesite maintainers]
+---
+
+# Snapshot social embeds into the owner's repo instead of rendering them remotely
+
+## Context and Problem Statement
+
+Owners want to embed content from social platforms — a tweet quoted in a blog post, the
+post a reply is replying to. ADR-0008 blocks the normal way of doing this: every platform's
+embed is a `<script>` from the platform's own CDN that tracks every visitor who loads the
+page, whether or not they interact with it.
+
+ADR-0008's stated alternative was "static screenshots with links to original posts". That
+is honest but manual, produces no machine-readable citation, and degrades the IndieWeb
+reply-context markup (`u-in-reply-to`, `h-cite`) that receiving Webmention endpoints parse.
+Anglesite needed a real embed story that keeps ADR-0008 intact.
+
+There is a second, less obvious problem. This is *legacy* social content: the premise is
+that these platforms are decaying. Any design that fetches from the platform at render time
+— or at build time — has bet the owner's archive on that platform's continued goodwill.
+X's oEmbed endpoint, Instagram's API, and the platforms themselves may not outlive the
+sites that cite them.
+
+## Decision Drivers
+
+* ADR-0008 — no third-party JavaScript, and no new exception if it can be avoided
+* Visitor privacy — a hotlinked image is still a tracked request, not just a script
+* ADR-0011 — the owner controls everything; the artifact must live in their repo
+* Git is the source of truth — the app must never be the only way to edit a site
+* Durability — an embed must survive the platform deleting the post or retiring its API
+* Build determinism — a deploy must not fail because a remote is down
+* IndieWeb correctness — reply context must emit parseable `h-cite` microformats
+
+## Considered Options
+
+* **Snapshot at author time into the owner's git repo** (chosen)
+* **Cloudflare Zaraz server-side rendering of embeds**
+* **`astro-embed` as shipped** (fetch at build time, hotlink media)
+* **Status quo** — static screenshots with links, per ADR-0008's original table
+
+## Decision Outcome
+
+Chosen: **snapshot at author time into the owner's git repo.**
+
+A CLI fetches the post once, when the owner adds it, and writes a normalized record to
+`src/embeds/<slug>.json` plus its media to `public/embeds/<slug>/`. Both are committed. At
+build time a Markdown plugin swaps a bare URL for a first-party card rendered from that
+committed record; the same renderer serves `inReplyTo` / `bookmarkOf` / `likeOf` reply
+context as `h-cite`. **The build makes no network request at all.**
+
+Consequences that follow directly:
+
+* The rendered card is first-party HTML and first-party images, so **no CSP allowlist entry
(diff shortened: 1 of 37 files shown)


docs/decisions/0025-snapshotted-embeds.md
```diff
@@ -0,0 +1,147 @@
...
+answer, and it was the specific proposal this decision was asked to evaluate.
+
+It was rejected because it solves the problem **at request time, at the edge**, so the
+rendered artifact exists nowhere the owner can see, keep, or move:
+
+| | Zaraz SSR | Snapshot |
+|---|---|---|
+| Third-party JS on page | None | None |
+| Third-party image requests | Proxied via the owner's domain | Self-hosted |
+| Platform coverage | X + Instagram only | Any URL (Open Graph) + per-platform adapters |
```
davidwkeith, line 86: Nit: this row claims the Snapshot approach covers "Any URL (Open Graph) + per-platform adapters," a broader claim than Zaraz's "X + Instagram only" — but the "Rejected: status quo" section below (lines 109-112) says Instagram specifically *can't* be snapshotted automatically and falls back to a manual screenshot. So for Instagram, Snapshot doesn't actually beat Zaraz on coverage; it's a wash (manual escape hatch vs. Zaraz's SSR). Worth a one-line caveat here, e.g. "(Instagram excepted — see status-quo fallback below)" so the table doesn't read as a strict superset.

---

davidwkeith (commented):
Reviewed. This is a well-written, docs-only ADR addition — no code, no MCP/schema surface, no template changes.

Checks:
- New ADR-0025 follows the existing MADR front-matter convention (matches ADR-0023/0024) and the next-available number (0025) doesn't collide with anything on `main`.
- The amendment to ADR-0008's alternatives table is minimal and correctly scoped — it doesn't add a CSP exception, consistent with the ADR's own argument.
- `docs/decisions/README.md` index entry added correctly.
- The 34 vendored `agent-skills/*/references/docs/decisions/0008-no-third-party-javascript.md` copies are identical mechanical re-exports of the canonical file (confirmed via diff), and the `agent-skills-export` CI check has already passed, so the export isn't stale.
- Left one inline nit on a minor internal inconsistency in the Zaraz-vs-Snapshot comparison table regarding Instagram coverage — non-blocking.

No dependabot PRs are open right now, so nothing to auto-merge this cycle.

---

docs/decisions/0025-snapshotted-embeds.md
davidwkeith: Fixed in a23cc49. You're right that the row read as a strict superset while the status-quo section says the opposite. Rather than a parenthetical I made the row state it plainly — Instagram is a wash, not a win: it blocks automated requests and falls back to the manual screenshot, where Zaraz renders it automatically. That's the honest comparison, and it's the one place Zaraz is genuinely ahead.

davidwkeith: Correction: this did **not** actually land in the merge, despite my earlier reply.

The merge of #424 raced a push — GitHub recorded the PR head as `0bbadb9` and merged that, while `a23cc49` (the fix) had landed on the branch but wasn't picked up. `origin/docs/adr-privacy-preserving-embeds` still points at it; `main` does not contain it.

Re-landed cleanly off current main in #425.

PR #429, opened 2026-07-27
fix(server): give recordEdit's commit-tree a git identity

### LINKED ISSUE #428: recordEdit's commit-tree has no git identity, so every container overlay edit silently fails to commit
## Summary

Every overlay text edit applied through the container-backed dev pipeline (`apply_edit` → `recordEdit` in `server/edit-history.mjs`) fails to produce a commit on the hidden `anglesite/edits` branch, because `git commit-tree` has no git identity available in the guest environment. `recordEdit` swallows the failure and returns `undefined` by design, so the `apply_edit` MCP reply never carries a `commit` field — which means Anglesite-app's entire persist-to-`Source/` pipeline (issue Anglesite/Anglesite-app#718, fixed app-side by Anglesite-app#737/#988/#1000) never even gets invoked. The edit renders correctly in the live preview but is silently lost on container teardown, reproducing #718's exact user-visible symptom via a different, earlier root cause.

## Repro / evidence

Confirmed via a full manual end-to-end smoke test against a real booted local `LocalContainerSiteRuntime` container in the Anglesite-app (Debug build, ad-hoc signed, both `run-container-probe.sh echo`/`boot` preflight checks passing):

1. Imported a fresh fixture site, opened it, let the container boot.
2. Double-clicked a heading in the preview, edited the text, clicked away. Preview re-rendered correctly (`apply_edit` succeeded and the guest's on-disk file changed).
3. Checked the host `Source/` git repo: `git log` unchanged, working tree clean — the edit never reached the host.
4. Anglesite-app's debug pane showed **zero** persist-related log lines (no `container:<siteID>:persist` source, no `git:identity` source) — confirming `LocalContainerSiteRuntime.persistEdit(commit:)` was never even called, which only happens when the `apply_edit` reply's `commit` field is nil (see `MCPApplyEditRouter.swift`'s `editPersister` closure, and the `On edit does not fire when reply has no commit` test case).

To isolate the guest-side cause, I reproduced the container's exact base image + execution shape directly:

```sh
container run --rm node:24-bookworm-slim bash -c '
  apt-get update -qq && apt-get install -y --no-install-recommends git -qq
  mkdir -p /tmp/testrepo && cd /tmp/testrepo && git init -q
  echo hello > file.txt && git add file.txt
  unset GIT_AUTHOR_NAME GIT_AUTHOR_EMAIL GIT_COMMITTER_NAME GIT_COMMITTER_EMAIL
  git commit -m initial; echo RC=$?
'
```

Result (git 2.39.5, running as root, no `~/.gitconfig`, no env identity — i.e. exactly this image's guest environment):

```
fatal: unable to auto-detect email address (got 'root@<hostname>.(none)')
RC=128
```

`recordEdit`'s bare `git commit-tree` call hits this same failure, caught by its outer `catch { return undefined; }` (by design — see the doc comment: "Returns the new commit SHA, or `undefined` on any failure"). Nothing in the container image (`container/Dockerfile`, `Containers/anglesite-dev/Dockerfile` in Anglesite-app), `entrypoint.sh`, or `edit-history.mjs` ever configures a git identity for the guest.

## Root cause

`server/edit-history.mjs`'s `recordEdit` calls `commit-tree` with no environment override:

```js
const commit = runGit(projectRoot, [
  "commit-tree", tree, "-p", parent, "-m", formatMessage({ file, files, range, message }),
]);
```

Compare `server/undo-edit.mjs` (lines ~102-110), which gets this right for its own `commit-tree` call:

```js
const env = {
  GIT_AUTHOR_NAME: "Anglesite",
  GIT_AUTHOR_EMAIL: "edits@anglesite.local",
  GIT_COMMITTER_NAME: "Anglesite",
  GIT_COMMITTER_EMAIL: "edits@anglesite.local",
};
const newCommit = runGit(projectRoot, ["commit-tree", parentTree, "-p", head, "-m", message], env);
```

`edit-history.mjs`'s `recordEdit` is missing the equivalent four lines on its own `commit-tree` call.

## Fix

Pass the same `GIT_AUTHOR_NAME`/`GIT_AUTHOR_EMAIL`/`GIT_COMMITTER_NAME`/`GIT_COMMITTER_EMAIL` env (matching `undo-edit.mjs`'s existing convention) to `recordEdit`'s `commit-tree` call in `server/edit-history.mjs`.

## Acceptance

A container-booted overlay text edit produces a `commit` field on the `apply_edit` reply (verifiable via `recordEdit`'s own return value, or end-to-end via Anglesite-app's `persistEdit` actually landing a commit in host `Source/`).

## Relates to

Anglesite/Anglesite-app#718, Anglesite/Anglesite-app#737, Anglesite/Anglesite-app#988, Anglesite/Anglesite-app#1000

davidwkeith:
## Summary

- `recordEdit`'s `commit-tree` call in `server/edit-history.mjs` had no `GIT_AUTHOR_*`/`GIT_COMMITTER_*` env, so it depended entirely on ambient git config/env for an identity.
- In the container-backed dev pipeline's guest environment (no `~/.gitconfig`, no identity env vars), `commit-tree` fails with `fatal: unable to auto-detect email address`, `recordEdit`'s catch-all swallows it and returns `undefined`, and `apply_edit`'s reply omits `commit` — so Anglesite-app's persist-to-`Source/` pipeline never fires and the edit is silently lost on container teardown.
- Fix: pass the same `GIT_AUTHOR_NAME`/`GIT_AUTHOR_EMAIL`/`GIT_COMMITTER_NAME`/`GIT_COMMITTER_EMAIL` (`Anglesite` / `edits@anglesite.local`) that `undo-edit.mjs` already uses for its own `commit-tree` call, so `recordEdit` never depends on ambient identity.

## Paired PR check

- [x] This change is **self-contained** to the plugin (MCP server fix with no consumer-visible contract change — `apply_edit`'s reply schema already had an optional `commit` field; this just makes it populate reliably in identity-less guest environments instead of silently omitting it).
- [ ] This change **needs a paired PR** in [`Anglesite/Anglesite-app`](https://github.com/Anglesite/Anglesite-app)

## Test plan

- [x] Added a regression test (`test/edit-history.test.js`) that reproduces the container guest exactly: no repo-local `user.name`/`user.email`, `GIT_CONFIG_GLOBAL=/dev/null` + `GIT_CONFIG_NOSYSTEM=1` to strip this sandbox's own global config, and no `GIT_AUTHOR_*`/`GIT_COMMITTER_*` env — verified it fails on the pre-fix code (`sha` is `undefined`) and passes with the fix.
- [x] Full `edit-history.test.js` + `undo-edit.test.js` suites pass (17/17).

Fixes #428

---
_Generated by [Claude Code](https://claude.ai/code/session_01ToF6XqGGxVVjzVSpQ4RJf8)_

### DIFF
--- a/server/edit-history.mjs
+++ b/server/edit-history.mjs
@@ -20,6 +20,7 @@ import { execFileSync } from "node:child_process";
 import { mkdtempSync, rmSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
+import { ANGLESITE_COMMIT_IDENTITY } from "./git-identity.mjs";
 
 const EDITS_REF = "refs/heads/anglesite/edits";
 
@@ -93,15 +94,9 @@ export async function recordEdit(projectRoot, { file, files, range, message }) {
       }
       const tree = runGit(projectRoot, ["write-tree"], env);
       const parent = runGit(projectRoot, ["rev-parse", EDITS_REF]);
-      const commitEnv = {
-        GIT_AUTHOR_NAME: "Anglesite",
-        GIT_AUTHOR_EMAIL: "edits@anglesite.local",
-        GIT_COMMITTER_NAME: "Anglesite",
-        GIT_COMMITTER_EMAIL: "edits@anglesite.local",
-      };
       const commit = runGit(projectRoot, [
         "commit-tree", tree, "-p", parent, "-m", formatMessage({ file, files, range, message }),
-      ], commitEnv);
+      ], ANGLESITE_COMMIT_IDENTITY);
       // CAS update with OLDVALUE so two concurrent recordEdits can't silently clobber each other.
       runGit(projectRoot, ["update-ref", EDITS_REF, commit, parent]);
       return commit;
--- a/server/edit-history.mjs
+++ b/server/edit-history.mjs
@@ -93,9 +93,15 @@ export async function recordEdit(projectRoot, { file, files, range, message }) {
       }
       const tree = runGit(projectRoot, ["write-tree"], env);
       const parent = runGit(projectRoot, ["rev-parse", EDITS_REF]);
+      const commitEnv = {
+        GIT_AUTHOR_NAME: "Anglesite",
+        GIT_AUTHOR_EMAIL: "edits@anglesite.local",
+        GIT_COMMITTER_NAME: "Anglesite",
+        GIT_COMMITTER_EMAIL: "edits@anglesite.local",
+      };
       const commit = runGit(projectRoot, [
         "commit-tree", tree, "-p", parent, "-m", formatMessage({ file, files, range, message }),
-      ]);
+      ], commitEnv);
       // CAS update with OLDVALUE so two concurrent recordEdits can't silently clobber each other.
       runGit(projectRoot, ["update-ref", EDITS_REF, commit, parent]);
       return commit;
--- a/server/git-identity.mjs
+++ b/server/git-identity.mjs
@@ -1,0 +1,15 @@
+/**
+ * Shared git identity for commits Anglesite itself authors on the hidden
+ * `refs/heads/anglesite/edits` branch (`edit-history.mjs`'s `recordEdit` and
+ * `undo-edit.mjs`'s `undoEdit`). Both call `commit-tree` directly, which — unlike
+ * `git commit` — never falls back to a "system default" identity: it needs
+ * GIT_AUTHOR_* / GIT_COMMITTER_* env (or matching git config) or it fails outright
+ * with "unable to auto-detect email address" in guest environments that have
+ * neither (see #428). Centralized here so the two callers can't drift.
+ */
+export const ANGLESITE_COMMIT_IDENTITY = {
+  GIT_AUTHOR_NAME: "Anglesite",
+  GIT_AUTHOR_EMAIL: "edits@anglesite.local",
+  GIT_COMMITTER_NAME: "Anglesite",
+  GIT_COMMITTER_EMAIL: "edits@anglesite.local",
+};
--- a/server/undo-edit.mjs
+++ b/server/undo-edit.mjs
@@ -16,6 +16,7 @@
 import { execFileSync } from "node:child_process";
 import { writeFileSync, unlinkSync } from "node:fs";
 import { join } from "node:path";
+import { ANGLESITE_COMMIT_IDENTITY } from "./git-identity.mjs";
 
 const EDITS_REF = "refs/heads/anglesite/edits";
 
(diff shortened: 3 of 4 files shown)


server/edit-history.mjs
davidwkeith, line 1: Minor: this identity block (`GIT_AUTHOR_NAME: \"Anglesite\"`, `edits@anglesite.local`, etc.) is now duplicated verbatim in `server/undo-edit.mjs` (its own `commit-tree` call already used the same four values). Worth pulling into a shared `ANGLESITE_COMMIT_IDENTITY` constant in a small shared module (or exporting it from one of these two files) so the two `commit-tree` callers can't drift if this identity ever changes. Not blocking — the fix itself is correct and well-tested.

---

davidwkeith (commented):
Solid, narrowly-scoped fix. `recordEdit`'s `commit-tree` call now gets an explicit identity, mirroring the same four env vars `undo-edit.mjs` already uses — so `commit-tree` no longer depends on ambient `~/.gitconfig`/env in identity-less environments (the exact container-guest failure from #428). The new regression test correctly reproduces that environment (`GIT_CONFIG_GLOBAL=/dev/null`, `GIT_CONFIG_NOSYSTEM=1`, no `GIT_AUTHOR_*`/`GIT_COMMITTER_*`) and would have failed pre-fix.

One small note left inline (non-blocking): the identity literal is now duplicated between `edit-history.mjs` and `undo-edit.mjs` — worth a shared constant at some point to avoid drift.

CI is still running as of this review but the change is small and well-tested; looks mergeable once green (note: I can't formally approve since this PR is under my own account, but no functional issues found).

---

davidwkeith (commented):
Looks correct. I checked `runGit` in `server/edit-history.mjs` — it does `env: { ...process.env, ...env }`, so passing `commitEnv` only overrides the four `GIT_AUTHOR_*`/`GIT_COMMITTER_*` keys and doesn't clobber anything else (`GIT_INDEX_FILE`, etc.) the surrounding calls rely on. The regression test does a thorough job isolating ambient identity (config unset, `GIT_CONFIG_GLOBAL=/dev/null`, `GIT_CONFIG_NOSYSTEM=1`, env vars cleared and restored in a `finally`), and it reproduces #428 exactly.

One minor nit, non-blocking: the identity literals (`"Anglesite"` / `"edits@anglesite.local"`) are now duplicated verbatim in both `server/undo-edit.mjs` and `server/edit-history.mjs`. Since both call sites need to agree for `undo_edit` to correctly attribute/revert `recordEdit`'s commits, consider hoisting them into one shared constant (e.g. a small `server/git-identity.mjs` or an export from one of the two files) so they can't drift apart in a future edit.

---

server/edit-history.mjs
davidwkeith: Good catch — hoisted into `server/git-identity.mjs` as `ANGLESITE_COMMIT_IDENTITY`, now imported by both `edit-history.mjs` and `undo-edit.mjs` (fe4a445).

---

